PR #1, opened 2015-01-12
Initial work

wprice:
Redis work to be checked into master at some point.


README.md
```diff
@@ -1,4 +1,66 @@
...
+
+The Vert.x Redis service provides asynchronous connectivity and  to a [Redis](http://redis.io) data-structure server. While API calls ultimately are against the Vert.x Event Bus [Vert.x](http://vertx.io), the service provides a rich API that allows you to develop using a the native idioms and vernacular of the Redis runtime. 
+
+The Vert.x Redis service is developed for use with Vert.x V3 and as such, is not backward compatible with a prior Vert.x release. Simiarly, the underlying Redis service was developed and testing with the latest stable release of Redis, which at the time of this writing was verstion [2.8.19](http://redis.io/download). 
+
+Keep in mind that while the Vert.x service provides support for the complete [Redis command set](http://redis.io/commands), a working knowledge of the Redis runtime is strongly encouraged. Where there appears to be a discrepancy between this documentation and the [Redis documentation set](http://redis.io/documentation), the Redis literature takes precedence. 
+
+This project is based on original work from Paulo Lopes. The original repository for the module can be found at [https://github.com/vert-x/mod-redis](http://).
```
nscavell: Missing the actual link

README.md
```diff
@@ -1,4 +1,66 @@
...
+Keep in mind that while the Vert.x service provides support for the complete [Redis command set](http://redis.io/commands), a working knowledge of the Redis runtime is strongly encouraged. Where there appears to be a discrepancy between this documentation and the [Redis documentation set](http://redis.io/documentation), the Redis literature takes precedence. 
+
+This project is based on original work from Paulo Lopes. The original repository for the module can be found at [https://github.com/vert-x/mod-redis](http://). 
+
+##Redis Service
+While it is assumed you are familar with Redis, using the Redis service is straightforward
+
+###Creating and Starting
+
+<pre>
```
nscavell: Instead of using \<pre\> you can use \`\`\` for code, and can also tell what language it is for syntax highlighting. i.e. \`\`\`java

README.md
wprice: Good catch. Will fix.

README.md
```diff
@@ -1,4 +1,66 @@
...
+    service.start(asyncResult -> {
+      if (asyncResult.succeeded()) {
+        String address = config().getString("address", "vertx.redis");
+        ProxyHelper.registerService(RedisService.class, vertx, service, address);
+        startFuture.complete();
+      } else {
+        startFuture.fail(asyncResult.cause());
+      }
+    });	
+
```
nscavell: I think it would be better to show deploying the verticle via the "service:" prefix and calling createEventBusProxy to obtain the actual service.

wprice: Yep, good one. Will. fix.

README.md
```diff
@@ -1,4 +1,66 @@
...
+    service.set(toJsonArray(mykey, "Hello"), reply0 -> {
+      assertTrue(reply0.succeeded());
+      redis.get(toJsonArray(mykey), reply1 -> {
+        assertTrue(reply1.succeeded());
+        assertEquals("Hello", reply1.result());
+        testComplete();
+      });
+    });
+    await();
+
```
nscavell: Instead of test code (not sure what toJsonArray and mykey is) I think a simple example someone can copy and paste would be good.

wprice: Yep, agreed.

pom.xml
```diff
@@ -0,0 +1,231 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>io.vertx</groupId>
+    <artifactId>vertx-parent</artifactId>
+    <version>3.0.0-SNAPSHOT</version>
+  </parent>
```
nscavell: You can inherit from vertx-ext-parent and reduce a lot of 'pom noise' and just include the service proxy stuff. Take a look at the jdbc service.

pom.xml
```diff
@@ -0,0 +1,231 @@
...
+    </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+      <version>4.11</version>
+    </dependency>
+  </dependencies>
+
+  <build>
```
nscavell: I think we can get rid of this entire build section when we inherit from ext-parent.

src/main/java/io/vertx/redis/RedisService.java
```diff
@@ -0,0 +1,1634 @@
...
+   */
+  void commandCount(Handler<AsyncResult<Long>> handler);
+
+  /**
+   * Extract keys given a full Redis command
+   *
+   * @since 2.8.13
+   * group: server
+   */
+  void commandGetkeys(Handler<AsyncResult<JsonArray>> handler);
```
nscavell: No input ?

wprice: No. There are a few of these in Redis that simple are commands where no input is required and simply return a result.

src/main/java/io/vertx/redis/RedisService.java
```diff
@@ -0,0 +1,1634 @@
...
+import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+import io.vertx.redis.impl.RedisServiceImpl;
+import io.vertx.serviceproxy.ProxyHelper;
+
+@VertxGen
+@ProxyGen
+public interface RedisService {
+
```
nscavell: So my issue with this approach (wrapping all args in a JsonArray) is the service API isn't really buying me much as apposed to sending it manually over the event bus. And with redis being so simple (for most commands) get and set for example should be pretty simple.

src/main/java/io/vertx/redis/RedisService.java
nscavell: Yea I was reading the redis docs incorrectly :)

src/main/java/io/vertx/redis/impl/AbstractRedisService.java
```diff
@@ -0,0 +1,245 @@
...
+
+  @Override
+  public void start(final Handler<AsyncResult<Void>> handler) {
+    final String host = config.getString("host", "localhost");
+    final int port = config.getInteger("port", 6379);
+    final String enc = config.getString("encoding", "UTF-8");
+    final boolean binary = config.getBoolean("binary", false);
+
+    if (binary) {
+      log.warn("Binary mode is not implemented yet!!!");
```
nscavell: Should we fail here ?

src/main/java/io/vertx/redis/impl/AbstractRedisService.java
```diff
@@ -0,0 +1,245 @@
...
+    final ResponseTransform transform = getResponseTransformFor(command);
+
+    // subscribe/psubscribe and unsubscribe/punsubscribe commands can have multiple (including zero) replies
+    int expectedReplies = 1;
+
+    switch (command) {
+      // argument "pattern" ["pattern"...]
+      case "psubscribe":
+        // in this case we need also to register handlers
+        if (redisArgs == null) {
```
nscavell: Check the size too since it can have 0 elements.

src/main/java/io/vertx/redis/impl/AbstractRedisService.java
```diff
@@ -0,0 +1,245 @@
...
+            message.put("channel", replyData[2].asType(String.class, encoding));
+            message.put("message", replyData[3].asType(String.class, encoding));
+            replyMessage.put("value", message);
+            eb.send(vertxChannel, replyMessage);
+          });
+        }
+        break;
+      // argument "channel" ["channel"...]
+      case "subscribe":
+        if (redisArgs == null) {
```
nscavell: Again, check size as well.

src/main/java/io/vertx/redis/impl/AbstractRedisService.java
```diff
@@ -0,0 +1,245 @@
...
+                // entry in section
+                int split = line.indexOf(':');
+                if (section == null) {
+                  value.put(line.substring(0, split), line.substring(split + 1));
+                } else {
+                  section.put(line.substring(0, split), line.substring(split + 1));
+                }
+              }
+            }
+            resultHandler.handle(new RedisAsyncResult<>(null, (T) value));
```
nscavell: I believe we can get rid of RedisAsyncResult and use Future.succeededFuture or Future.failedFuture

src/main/java/io/vertx/redis/spi/RedisServiceFactory.java
```diff
@@ -0,0 +1,12 @@
+package io.vertx.redis.spi;
+
+import io.vertx.core.Vertx;
+import io.vertx.core.json.JsonObject;
+import io.vertx.redis.RedisService;
+
+public interface RedisServiceFactory {
```
nscavell: I believe we can get rid of this

src/main/java/io/vertx/redis/package-info.java
```diff
@@ -0,0 +1,3 @@
+@GenModule(name = "redis") package io.vertx.redis;
```
nscavell: Module name should be vertx-redis

src/main/java/io/vertx/redis/impl/AbstractRedisService.java
wprice: Probably not as the service is still usable, the binary protocol just isn't supported.It's not as if they have misconfigured a host name or port or something requiring a complete failure.

PR #3, opened 2015-04-14
Make RedisService more type-safe

mstruk:
- a great number of methods was refactored, 
- additional tests were added
- some tests were fixed to avoid intermittent failures - hopefully none of the fixes were properly failing, and implementation should be fixed rather then the test
- some tests have been commented out, and have to be reenabled, and fixed
- additional build profile was added for testing against standalone latest Redis version
- Redis connection lifecycle needs to be looked at more - specifically how one should use the service in situations when a set of commands has to be executed as part of the same client connection, and if current implementation properly support such a usecase


purplefox:
Thanks a lot for this. It is a great improvement.

Issue #12, opened 2015-06-03
add missing commands


pmlopes:
Most of the 3.x version commands are missing (these regard to cluster management mostly)

pmlopes:
fixed on master

PR #14, opened 2015-06-29
#12 add missing cluster commands

pmlopes:
Add missing 3.0 API related to clustering, testing however is more complicated since both docker images and embedded redis do not support clustering, still need to figure out how to automate testing for this.

There is nothing really exciting here just add the commands to the client interface and delegate to the generic redis command on the impl.

@purplefox @vietj @cescoffier please review

### DIFF
--- a/src/main/java/io/vertx/redis/RedisClient.java
+++ b/src/main/java/io/vertx/redis/RedisClient.java
@@ -241,6 +241,216 @@ public interface RedisClient {
   RedisClient clientSetname(String name, Handler<AsyncResult<String>> handler);
 
   /**
+   * Assign new hash slots to receiving node.
+   *
+   * @param slots
+   * @param handler Handler for the result of this call.
+   * @since 3.0.0
+   * group: server
+   */
+  RedisClient clusterAddslots(List<Long> slots, Handler<AsyncResult<Void>> handler);
+
+  /**
+   * Return the number of failure reports active for a given node.
+   *
+   * @param nodeId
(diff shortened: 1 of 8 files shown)


src/main/java/io/vertx/redis/RedisClient.java
```diff
@@ -241,6 +241,216 @@ static RedisClient create(Vertx vertx, JsonObject config) {
...
+
+  /**
+   * Return the number of local keys in the specified hash slot.
+   *
+   * @param slot
+   * @param handler Handler for the result of this call.
+   * @since 3.0.0
+   * group: cluster
+   */
+  RedisClient clusterCountkeysinslot(long slot, Handler<AsyncResult<Long>> handler);
```
purplefox: Shouldn't it be clusterCountKeysInSlot (i.e. camel case) ?

purplefox:
Looks ok, my only comment would be with method naming.

I think we should use camel case, e.g.

Not "clusterSetslotWithNode" but "clusterSetSlotWithNode".

pmlopes:
I did not respect the camel case rules to stay aligned with the existing API, currently the camel humps are only applied on the space between redis commands, for example we have `slowlog`, `zrangebylex` etc because those are the redis commands and in this case: `clusterCountkeysinslot` is `cluster countkeysinslot` i can fix the casing here but the remaining API probably i can't fix it now since it will break backwards compatibility.

purplefox:
Ok fair enough. Looks good as is then.

PR #16, opened 2015-06-30
convert commands to lowercase to compare with switch cases string literals

alanma:
The switch case comparison fails for command as of 3.0.0 since we pass ing upper case commands, but in the switch cases we're using lower case string literals.


pmlopes:
Thanks!

Issue #17, opened 2015-07-01
Weird problem when sending request to a basic Vert.x server with a basic Redis client.


poiuytrez:
I am currently working on a project where I need to receive an id number through a requests and then look if it is stored in my Redis database.

My project works well for one request or two, but when testing it with Jmeter and sending 3000 requests I noticed a weird problem.

I made a way more basic project and I published it to github to understand what was going on. 

There is actually two problems : 

- The first one being that sometimes my project won't respond to Jmeter after like responding 200 requests or so... Jmeter keeps waiting for the reply.

- The other one is that sometimes I get this exception, it has the same effect on Jmeter than the first one.

io.vertx.core.impl.ContextImpl
SEVERE: Unhandled exception
java.lang.RuntimeException: Received a non pub/sub message without reply handler waiting:io.vertx.redis.impl.Reply@6a3c7ffb
at io.vertx.redis.impl.RedisConnection.handleReply(RedisConnection.java:92)
at io.vertx.redis.impl.ReplyParser.handle(ReplyParser.java:154)
at io.vertx.redis.impl.ReplyParser.handle(ReplyParser.java:6)
at io.vertx.core.net.impl.NetSocketImpl.handleDataReceived(NetSocketImpl.java:309)
at io.vertx.core.net.impl.VertxNetHandler.lambda$channelRead$27(VertxNetHandler.java:54)
at io.vertx.core.net.impl.VertxNetHandler$$Lambda$47/140297772.run(Unknown Source)
at io.vertx.core.impl.ContextImpl.lambda$wrapTask$15(ContextImpl.java:312)
at io.vertx.core.impl.ContextImpl$$Lambda$6/60559178.run(Unknown Source)
at io.vertx.core.impl.ContextImpl.executeFromIO(ContextImpl.java:217)
at io.vertx.core.net.impl.VertxNetHandler.channelRead(VertxNetHandler.java:54)
at io.vertx.core.net.impl.VertxNetHandler.channelRead(VertxNetHandler.java:31)
at io.vertx.core.net.impl.VertxHandler.channelRead(VertxHandler.java:124)
at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:308)
at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:294)
at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:846)
[...]

PR #18, opened 2015-07-01
Feature/issue 17

pmlopes:
Fixes running on right context for all handlers since with load this causes issues (see #17)


pmlopes:
@purplefox @vietj @cescoffier could you review, basically all haldlers are now called inside vertx context, running the given test case from #17 using apache bench now always works while before aprox 1 in 1000 to 10000 times would fail.

src/main/java/io/vertx/redis/impl/RedisConnection.java
```diff
@@ -19,45 +17,70 @@
...
+    this.vertx = vertx;
     this.host = host;
     this.port = port;
     this.subscriptions = subscriptions;
-    this.closedHandler = closedHandler;
+    this.replyParser = new ReplyParser(vertx, this);
     this.client = vertx.createNetClient(new NetClientOptions());
   }
 
+  State state() {
```
cescoffier: may need synchronized (need to double check)

src/main/java/io/vertx/redis/impl/AbstractRedisClient.java
```diff
@@ -260,62 +267,46 @@ final void sendJsonObject(final String command, final JsonArray args, final Hand
...
-    final String host = config.getString("host", "localhost");
-    final int port = config.getInteger("port", 6379);
-    RedisConnection conn = new RedisConnection(vertx, host, port, subscriptions, v -> connectionClosed());
-    conn.connect(res -> {
+    redisConnection.connect(res -> {
       if (res.succeeded()) {
-        connected(conn);
+        Command command;
+        while ((command = pending.poll()) != null) {
+          redisConnection.send(command);
```
cescoffier: may the connection be closed in the middle of the processing ?

src/main/java/io/vertx/redis/impl/RedisConnection.java
```diff
@@ -19,45 +17,70 @@
...
-      if (asyncResult.failed()) {
-        resultHandler.handle(Future.failedFuture(asyncResult.cause()));
-      } else {
-        netSocket = asyncResult.result();
-        netSocket.handler(replyParser);
-        netSocket.closeHandler(closedHandler);
-        resultHandler.handle(Future.succeededFuture());
-      }
-    });
+    if (state == State.DISCONNECTED) {
```
cescoffier: does this require a synchronization ?

src/main/java/io/vertx/redis/impl/RedisConnection.java
```diff
@@ -19,45 +17,70 @@
...
   private final int port;
-  private final Handler<Void> closedHandler;
 
+  public enum State {
+    DISCONNECTED,
+    CONNECTING,
+    CONNECTED
+  }
+
+  private State state = State.DISCONNECTED;
```
cescoffier: should it be declared volatile ?

cescoffier:
I've opened a couple of questions about the synchronisation. Everything else is ok.

src/main/java/io/vertx/redis/impl/AbstractRedisClient.java
```diff
@@ -115,7 +113,8 @@ final void sendJsonObject(final String command, final JsonArray args, final Hand
       case "psubscribe":
         // in this case we need also to register handlers
         if (redisArgs == null) {
-          resultHandler.handle(new RedisAsyncResult<>("at least one pattern is required!"));
+          vertx.getOrCreateContext().runOnContext(v ->
```
purplefox: Instead of calling vertx.getOrCreateContext() multiple times in the same method, it would be nicer to call it once at the top and just use the context obtained later on.

src/main/java/io/vertx/redis/impl/AbstractRedisClient.java
```diff
@@ -190,23 +190,24 @@ final void sendJsonObject(final String command, final JsonArray args, final Hand
...
 
-    Context context = vertx.getOrCreateContext();
-
-    Command<T> comm = new Command<T>(context, command, redisArgs, binary ? binaryCharset : charset)
+    Command<T> comm = new Command<T>(command, redisArgs, binary ? binaryCharset : charset)
                      .setExpectedReplies(expectedReplies).setHandler((Reply reply) -> {
         switch (reply.type()) {
           case '-': // Error
-            resultHandler.handle(new RedisAsyncResult<>(reply.asType(String.class)));
+            vertx.getOrCreateContext().runOnContext(v ->
```
purplefox: This doesn't look right to me. When this is invoked it will be on the context of the netsocket of the connection - that's probably going to be a different context to person who sent the original message.

I think it should invoked on the same context as he who sent the message.

pmlopes:
@cescoffier @purplefox after lots of rework i think the redis client is in good shape, it now supports pub/sub mode with test, recovery of closed connections works and there are tests, more tests were added in regards to transactions and the issue with the context has been fixed. I've tested it using the test example:

```
    JsonObject config = new JsonObject()
        .put("host", "127.0.0.1")
        .put("port", 6379);

    final RedisClient redis = RedisClient.create(Vertx.vertx(), config);

    HttpServer server = vertx.createHttpServer();

    server.requestHandler(req -> {
      redis.get("k", res -> {
        if (res.succeeded()) {
          req.response().setStatusCode(200).end();
        } else {
          System.out.println("ERR");
          req.response().setStatusCode(500).end();
        }
      });
    }).listen(8000);
```

And the runnning 4 instances of ApacheBench `ab -n 100000 -c 10 http://localhost:8000/` so all my cores were busy and the issue of handlers not being called back does not reproduce anymore, all requests were correctly handled.

The test was also repeated now using 4 verticle instances with the same successful results.

Please review since it is quite a big chunk of changes.

purplefox:
Ok, will review this

purplefox:
Looks good. My only comment is that it would be good to add a few tests that check to see that the response is handled in the correct context.

E.g. share a redis connection from different contexts and assert that the response context is correct.

Also check the case where the client is called from outside a context.

PR #32, opened 2015-09-14
Fixes #30 - Eval/Evalsha return type should be JsonArray not Void

### LINKED ISSUE #30: Eval not returning object type of result
In Jedis client, Eval op will return Object. But in redis-client here we return nothing with a void type. Any specific reason that we could not return object type? I could work on a PR if we want the eval and eval sha to be able to return object type data.

Thanks.

pmlopes:
Fixes #30

All eval commands are casted to a json array no matter what return type was present in order to have codegen to generate consistent API across languages


alanma:
@pmlopes the changes look solid for returning proper type pf result. I do have a question and suggestion while going through the code. Right now, the Reply.type is a byte, which I guess is to save some bytes to store the reply type. Other than that, is there other reasons that we could not use an enum to make the code more readable? Thanks.

alanma:
@pmlopes I think the changes for the result type is solid. 

One minor suggestion, do you think we could add an enum for the reply type to encapsulate the RESP protocol description(+, -, $, :, *)? This might make the code more readable.

Thanks.

pmlopes:
@alanma yes we can do that but lets do that in a separate pull request. Since those values are bytes maybe we could use a old interface with constants

pmlopes:
test this please

pmlopes:
@vietj @cescoffier @purplefox could you please review

cescoffier:
@pmlopes I've done a review. Almost fine for me, a couple of cosmetic comment.

PR #34, opened 2015-09-16
Reconnect will handle auth/select

### LINKED ISSUE #31: Redis client reconnect fails without AUTH
@pmlopes 

My redis server has AUTh required. So when I initialize my redis client, I have to manually call AUTH command after creating the redis client. Now I am seeing my redis-client losing connection to my redis server - not sure why yet - but reconnecting to the redis server does not call AUTH command. Therefore, queued commands won't be able to run after trying reconnecting with the exception:

`NOAUTH Authentication required.`

How should we reconnect to redis server requiring AUTH? I'd think we should call AUTH commands in the connect function in the RedisConnection class. Thoughts?

Thanks.

pmlopes:
Fixes #31

Currently the redis client reconnect to redis in case of socket errors, however if one was not using the default database or was connecting to a password protected server the reconnect would fail because handshake `auth` and/or `select` where not performed before resending all pending commands.

The client now does the full handshake. (reproducer tests included).


WindcodeFr:
Little error in `RedisOptions`.
```
/**
* Set the eventbus address prefix for `PUB/SUB`.
* @param address address prefix.
* @return self
*/
public RedisOptions setAddress(String address) {
    json.put("encoding", address);
    return this;
}
```

Must be ? =>
```
/**
* Set the eventbus address prefix for `PUB/SUB`.
* @param address address prefix.
* @return self
*/
public RedisOptions setAddress(String address) {
    json.put("address", address);
    return this;
}
```

alanma:
The implementation of doauth and doselect during the connection recovery looks correct to me. May we please merge it into master? :+1:

src/main/java/io/vertx/redis/impl/RedisConnection.java
```diff
@@ -279,6 +254,88 @@ void send(final Command<?> command) {
...
+    while ((command = pending.poll()) != null) {
+      // The order read must match the order written, vertx guarantees
+      // that this is only called from a single thread.
+      for (int i = 0; i < command.getExpectedReplies(); ++i) {
+        waiting.add(command);
+      }
+
+      // write to the socket in the netSocket context
+      final Command<?> _command = command;
+      context.runOnContext(v -> _command.writeTo(netSocket));
```
vietj: not sure this is needed, you should be able to write to a NetSocket outside of the context.

PR #37, opened 2015-09-28
ensure api is not broken with 3.x

pmlopes:
make sure that we don't break API within 3.x

### DIFF
--- a/src/main/java/io/vertx/redis/RedisClient.java
+++ b/src/main/java/io/vertx/redis/RedisClient.java
@@ -22,7 +22,7 @@ public interface RedisClient {
   }
 
   /**
-   * For type safety please use {@link #create(Vertx, RedisOptions)}.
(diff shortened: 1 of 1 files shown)


cescoffier:
Breaking changes have been listed here: https://github.com/vert-x3/wiki/wiki/3.1.0---Breaking-changes

src/main/java/io/vertx/redis/RedisClient.java
```diff
@@ -20,6 +21,15 @@ static RedisClient create(Vertx vertx) {
     return new RedisClientImpl(vertx, new RedisOptions());
   }
 
+  /**
+   * For type safety please use {@link #create(Vertx, RedisOptions)}.
+   */
```
cescoffier: Missing @deprecated in the javadoc:

@Deprecated For type safety please use {@link #create(Vertx, RedisOptions)}.

Issue #41, opened 2015-10-20
redisClient.setBinaryWithOptions method is missing (ref jedis)


RichardHightower:
```java

        final SetOptions setOptions = new SetOptions();
        setOptions.setPX(expiry.toMillis());

        redisClient.setBinaryWithOptions(key, value,  setOptions, event -> {
            if (event.failed()) {
                confirmation.onError(event.cause());
            } else {
                confirmation.accept(true);
            }
        });
```

pmlopes:
since we already have this API (although not a redis command) we should have the full implementation.

Note however that we do not use jedis or any other redis client underneath so do not assume that if an API exists on other drivers it should exist here.

RichardHightower:
I know you are not using jedis. Jedis has this because it makes sense. I was referring to prior art. 

Redis does not make a distinction between text and binary with the set method. 
Java, Scala, JavaScript, Groovy, etc. do.

This is why we need a version of setBinary and setBinaryWithOptions that take a Vertx Buffer so that we can set binary data into Redis. 

See Jedis for prior art.

RichardHightower:
nice

PR #43, opened 2015-10-21
Issue 40

### LINKED ISSUE #40: redisClient.setBinary should take a byte[] or a Buffer not a String (reference jedis)
```java

byte [] value ....
        redisClient.setBinary(key, value, event -> {
            if (event.failed()) {
                confirmation.onError(event.cause());
            } else {
                confirmation.accept(true);
            }
        });
```

pmlopes:
Fixes #40 
Depends on #41

### DIFF
--- a/src/main/java/io/vertx/redis/RedisClient.java
+++ b/src/main/java/io/vertx/redis/RedisClient.java
@@ -21,6 +21,7 @@ import io.vertx.codegen.annotations.VertxGen;
 import io.vertx.core.AsyncResult;
 import io.vertx.core.Handler;
 import io.vertx.core.Vertx;
+import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonArray;
 import io.vertx.core.json.JsonObject;
 import io.vertx.redis.impl.RedisClientImpl;
@@ -830,7 +831,7 @@ public interface RedisClient {
    * group: string
    */
   @Fluent
-  RedisClient getBinary(String key, Handler<AsyncResult<String>> handler);
+  RedisClient getBinary(String key, Handler<AsyncResult<Buffer>> handler);
 
   /**
    * Returns the bit value at offset in the string value stored at key
@@ -1833,7 +1834,7 @@ public interface RedisClient {
    * group: string
    */
   @Fluent
-  RedisClient setBinary(String key, String value, Handler<AsyncResult<Void>> handler);
+  RedisClient setBinary(String key, Buffer value, Handler<AsyncResult<Void>> handler);
 
   /**
    * Set the string value of a key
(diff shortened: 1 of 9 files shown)


pmlopes:
@cescoffier @vietj @purplefox could you review? For Binary payloads the API now uses Buffers and internally i don't do any conversions to/from json.

src/main/java/io/vertx/redis/RedisClient.java
```diff
@@ -830,7 +831,7 @@ static RedisClient create(Vertx vertx, RedisOptions config) {
    * group: string
    */
   @Fluent
-  RedisClient getBinary(String key, Handler<AsyncResult<String>> handler);
+  RedisClient getBinary(String key, Handler<AsyncResult<Buffer>> handler);
```
cescoffier: Breaking change - should be listed or avoided.

pmlopes: it is already listed ;)

cescoffier: Perfect !

src/main/java/io/vertx/redis/RedisClient.java
```diff
@@ -1833,7 +1834,20 @@ static RedisClient create(Vertx vertx, RedisOptions config) {
...
+  RedisClient setBinary(String key, Buffer value, Handler<AsyncResult<Void>> handler);
+
+  /**
+   * Set the string value of a key
+   *
+   * @param key     Key of which value to set
+   * @param value   New value for the key
+   * @param options Set options
+   * @param handler Handler for the result of this call.
+   * @since 1.0.0
```
cescoffier: I don't get the @since here. Aren't we in 3.2.0 ?

src/main/java/io/vertx/redis/RedisClient.java
```diff
@@ -1833,7 +1834,20 @@ static RedisClient create(Vertx vertx, RedisOptions config) {
...
+   * group: string
```
cescoffier: I don't get this either.

cescoffier:
Two small question about a javadoc comment. The rest looks good.

src/main/java/io/vertx/redis/RedisClient.java
pmlopes: that is the redis server api level all methods are annotated with the redis api server that supports it, it was something that was copied from the original client that i wrote. I can remove them all since the javadoc annotation does not refer to the API version

cescoffier: oh, well, I think it's a valuable information. Maybe instead of just @since, add something like @since Redis 1.0.0 to indicate it refers to Redis version and not the version of the current API.

pmlopes:
Sure it is a simple find/replace

PR #51, opened 2016-03-09
Fixes: #50 - setWithOptions does return simple String value

### LINKED ISSUE #50: setWithOptions shall return a result of String type
setWithOptions shall return a result of String type, install of Void. Otherwise it is no way to know the exact result when NX option is used.
SET command return string OK or (nil)

pmlopes:
Fix #50

### DIFF
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -1,5 +1,19 @@
 = Cheatsheets
 
+[[LimitOptions]]
+== LimitOptions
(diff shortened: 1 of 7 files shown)


cescoffier:
Thinking about compatibility. I guess it's ok, as before the user would not have used the parameter:

    ...,
   ar -> { 
      // do nothing with ar.result()
   }

 Now he can use it. The only weird case would be a check to `null`, but I guess this is more a bug in his application.

So :+1:

Issue #52, opened 2016-03-19
java.lang.ClassCastException, While using hgetall in a transaction


Ashwin-Surana:
Snippet code:
```
client = RedisClient.create(vertx, new RedisOptions().setAddress("127.0.0.1").setPort(6379));
client.multi(multiEvent -> {
  client.hgetall("hash1", hgetallEvent -> {
    client.exec(execEvent -> /* DO ANYTHING */);
  });
});
```
hgetall command inside a transaction always generates a CastException. 
```
SEVERE: Unhandled exception
java.lang.ClassCastException: java.lang.String cannot be cast to [Lio.vertx.redis.impl.Reply;
    at io.vertx.redis.impl.Reply.asType(Reply.java:124)
    at io.vertx.redis.impl.Reply.asType(Reply.java:156)
    at io.vertx.redis.impl.RedisConnection.handleReply(RedisConnection.java:376)
    at io.vertx.redis.impl.ReplyParser.handle(ReplyParser.java:180)
    at io.vertx.redis.impl.ReplyParser.handle(ReplyParser.java:21)
    at io.vertx.core.net.impl.NetSocketImpl.handleDataReceived(NetSocketImpl.java:313)
    at io.vertx.core.net.impl.VertxNetHandler.lambda$channelRead$30(VertxNetHandler.java:54)
    at io.vertx.core.impl.ContextImpl.lambda$wrapTask$18(ContextImpl.java:333)
    at io.vertx.core.impl.ContextImpl.executeFromIO(ContextImpl.java:225)
    at io.vertx.core.net.impl.VertxNetHandler.channelRead(VertxNetHandler.java:54)
    at io.vertx.core.net.impl.VertxNetHandler.channelRead(VertxNetHandler.java:31)
    at io.vertx.core.net.impl.VertxHandler.channelRead(VertxHandler.java:124)
    at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:318)
    at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:304)
    at io.netty.channel.DefaultChannelPipeline.fireChannelRead(DefaultChannelPipeline.java:846)
    at io.netty.channel.nio.AbstractNioByteChannel$NioByteUnsafe.read(AbstractNioByteChannel.java:131)
    at io.netty.channel.nio.NioEventLoop.processSelectedKey(NioEventLoop.java:511)
[...]

PR #54, opened 2016-03-29
Fix for issue #52 ( java.lang.ClassCastException, While using hgetall in a transaction)

Ashwin-Surana:
The issue #52 was that all the commands in Redis MULTI will return a string as "QUEUED" or syntax error if any(src: navigate to "Errors inside a Transaction" in this [site](http://redis.io/topics/transactions) ) . So while using hgetall inside transaction, it was trying to cast String - "QUEUED" to JsonObject which generated the CastException. I have implemented Transaction as an entity. Here is an [example-code](https://github.com/Ashwin-Surana/vertx-redis-client/blob/transaction-implementation/src/main/java/examples/Examples.java) on how to use this transaction entity (Check example5() method). This fix was achieved by removing multi, exec, discard, watch and unwatch methods from RedisClient and making them available in [RedisTransaction](https://github.com/Ashwin-Surana/vertx-redis-client/blob/transaction-implementation/src/main/java/io/vertx/redis/RedisClient.java) (Line 53 - nested in RedisClient) with other Commands accepting Handlers to be `Handlers<AsyncResult<String>>`.
Tests involving transactions were changed to this approach and all the tests passed. I am free to make the necessary changes for the fix to be accepted.

### DIFF
--- a/src/main/java/examples/Examples.java
+++ b/src/main/java/examples/Examples.java
@@ -5,10 +5,7 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.redis.RedisClient;
 import io.vertx.redis.RedisOptions;
 
-import java.util.Arrays;
-
 /**
- *
  * These are the examples used in the documentation.
  *
  * @author <a href="mailto:pmlopes@gmail.com">Paulo Lopes</a>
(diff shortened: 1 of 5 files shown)


pmlopes:
Thanks for the PR!

PR #55, opened 2016-04-03
bnd config to generate OSGi metadata

paulbakker:
Added bnd configuration to all auth providers to generate OSGi manifests. Test cases are available in a Bndtools workspace here: https://github.com/paulbakker/vertx-osgi-testcases/tree/master

Note that this requires the unreleased bnd-maven-plugin 3.2 plugin.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -130,6 +130,47 @@
           </execution>
         </executions>
       </plugin>
+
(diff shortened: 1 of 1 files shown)


cescoffier:
Could we integrate the test in the source code ?

PR #56, opened 2016-05-18
Implement Redis GEO API

### LINKED ISSUE #49: add missing geo api
http://redis.io/commands#geo

pmlopes:
Fixes #49

### DIFF
--- a/src/main/java/io/vertx/redis/op/GeoMember.java
+++ b/src/main/java/io/vertx/redis/op/GeoMember.java
@@ -1,0 +1,74 @@
+/**
+ * Copyright 2015 Red Hat, Inc.
+ *
+ *  All rights reserved. This program and the accompanying materials
+ *  are made available under the terms of the Eclipse Public License v1.0
+ *  and Apache License v2.0 which accompanies this distribution.
+ *
+ *  The Eclipse Public License is available at
+ *  http://www.eclipse.org/legal/epl-v10.html
+ *
+ *  The Apache License v2.0 is available at
+ *  http://www.opensource.org/licenses/apache2.0.php
+ *
+ *  You may elect to redistribute this code under either of these licenses.
+ */
+package io.vertx.redis.op;
+
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * @author <a href="mailto:plopes@redhat.com">Paulo Lopes</a>
+ */
+@DataObject
+public class GeoMember {
+
+  private Double longitude;
+  private Double latitude;
+  private String member;
+
+  public GeoMember() {}
+
+  public GeoMember(GeoMember obj) {
+    longitude = obj.longitude;
+    latitude = obj.latitude;
+    member = obj.member;
+  }
+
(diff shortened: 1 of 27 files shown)


src/main/java/io/vertx/redis/op/GeoMember.java
```diff
@@ -0,0 +1,74 @@
+/**
+ * Copyright 2015 Red Hat, Inc.
```
cescoffier: Wrong copyright holder ?

src/main/java/io/vertx/redis/op/GeoMember.java
```diff
@@ -0,0 +1,74 @@
...
+
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.json.JsonArray;
+import io.vertx.core.json.JsonObject;
+
+/**
+ * @author <a href="mailto:plopes@redhat.com">Paulo Lopes</a>
+ */
+@DataObject
+public class GeoMember {
```
cescoffier: Any reason to not use converters ?

src/main/java/io/vertx/redis/op/GeoUnit.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package io.vertx.redis.op;
+
+import io.vertx.codegen.annotations.VertxGen;
+
+/**
+ * @author <a href="mailto:plopes@redhat.com">Paulo Lopes</a>
+ */
+@VertxGen
+public enum GeoUnit {
```
cescoffier: Should have javadoc as it will create the data sheet for this enum

src/main/java/io/vertx/redis/op/GeoRadiusOptions.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  public GeoRadiusOptions(JsonObject obj) {
+    withcoord = obj.getBoolean("withcoord", Boolean.FALSE);
+    withdist = obj.getBoolean("withdist", Boolean.FALSE);
+    withhash = obj.getBoolean("withhash", Boolean.FALSE);
+    count = obj.getLong("count");
+  }
+
+  public GeoRadiusOptions setWithCoord(boolean bool) {
```
cescoffier: Fields should have javadoc to make the dat sheet more usable.

src/main/java/io/vertx/redis/op/GeoMember.java
```diff
@@ -0,0 +1,74 @@
...
+    member = obj.member;
+  }
+
+  public GeoMember(JsonObject obj) {
+    longitude = obj.getDouble("longitude");
+    latitude = obj.getDouble("latitude");
+    member = obj.getString("member");
+  }
+
+  public GeoMember setLongitude(Double val) {
```
cescoffier: Properties should be documented for the data sheet.

src/main/java/io/vertx/redis/op/GeoMember.java
pmlopes: The reason is that these DataObjects are never serialized to JSON, they are used to enforce types in the options but redis only works with Arrays, so what happens is that there is this `toJsonArray` where the real conversion happens.

PR #59, opened 2016-06-02
Fixes ClassCastException for SPOP many (it always MULTIBULK not BULK)

### LINKED ISSUE #58: ClassCastException, while using spopMany
### Version

* vert.x core:3.2.1
* vert.x redis client:3.2.1

<pre>
    redisClient.spopMany("FOO", 3, ar -> {
      if (ar.succeeded()) {
        System.out.println(ar.result());
      } else {
        System.out.println(ar.cause());
      }
    });
</pre>

the result of  ` SPOP key [count] `  should be JsonArray
exception:

> java.lang.ClassCastException: io.vertx.core.json.JsonArray cannot be cast to java.lang.String

pmlopes:
Fixes #58 

The interface was wrong since redis always returns a multibulk (list of strings) (broken API)

### DIFF
--- a/src/main/generated/io/vertx/rxjava/redis/RedisClient.java
+++ b/src/main/generated/io/vertx/rxjava/redis/RedisClient.java
@@ -4234,7 +4234,7 @@ public class RedisClient {
    * @param handler Handler for the result of this call.
    * @return 
    */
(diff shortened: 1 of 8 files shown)


cescoffier:
Looks good to me.

Must be listed on the breaking change page if not done already.

PR #62, opened 2016-07-01
Fix #61: synchronize around queue access and socket writes otherwise …

### LINKED ISSUE #61: java.lang.ClassCastException: java.lang.Long cannot be cast to [Lio.vertx.redis.impl.Reply;
_From @hansen2014 on June 28, 2016 9:58_

RedisTransaction rtx = redisClient.transaction();

		rtx.multi(r -> {

			String setKey = RepositoryHelper.createKSKey("SessionScreenSymbol", clientCd, sessionId, viewNum);

			List<String> keyList = new ArrayList<>();
			for (String data : delS) {
				keyList.add(data);
				rtx.del(data, r1 -> {
					GLog.info("trans del executing ..., result: {} ", r1);
				});
			}

			if (delS.size() > 0) {
				rtx.sremMany(setKey, keyList, rst -> {
					GLog.info("trans strm executing ..., result: {}", rst);
				});
			}

			rtx.exec(handler);
		});

sometimes, will throw exception like that , and then, can not receive the reply. : 
fatal: Unhandled exception
java.lang.ClassCastException: java.lang.Long cannot be cast to [Lio.vertx.redis.impl.Reply;
	at io.vertx.redis.impl.Reply.asType(Reply.java:94)
	at io.vertx.redis.impl.Reply.asType(Reply.java:156)
	at io.vertx.redis.impl.RedisConnection.handleReply(RedisConnection.java:441)
	at io.vertx.redis.impl.ReplyParser.handle(ReplyParser.java:180)
	at io.vertx.redis.impl.ReplyParser.handle(ReplyParser.java:21)
	at io.vertx.core.net.impl.NetSocketImpl.handleDataReceived(NetSocketImpl.java:312)
	at io.vertx.core.net.impl.VertxNetHandler.lambda$channelRead$0(VertxNetHandler.java:67)
	at io.vertx.core.impl.ContextImpl.lambda$wrapTask$3(ContextImpl.java:357)
	at io.vertx.core.impl.OrderedExecutorFactory$OrderedExecutor.lambda$new$0(OrderedExecutorFactory.java:94)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)
	at java.lang.Thread.run(Unknown Source)

2016-06-28 17:57:57.558 [vert.x-eventloop-thread-7] ERROR EAGLE.DEFAULT - system error.
io.vertx.core.eventbus.ReplyException: Timed out after waiting 5000(ms) for a reply. address: 136
	at io.vertx.core.eventbus.impl.HandlerRegistration.sendAsyncResultFailure(HandlerRegistration.java:118) ~[vertx-core-3.3.0.jar:na]
	at io.vertx.core.eventbus.impl.HandlerRegistration.lambda$new$0(HandlerRegistration.java:65) ~[vertx-core-3.3.0.jar:na]
	at io.vertx.core.impl.VertxImpl$InternalTimerHandler.handle(VertxImpl.java:769) ~[vertx-core-3.3.0.jar:na]
	at io.vertx.core.impl.VertxImpl$InternalTimerHandler.handle(VertxImpl.java:740) ~[vertx-core-3.3.0.jar:na]
	at io.vertx.core.impl.ContextImpl.lambda$wrapTask$3(ContextImpl.java:359) ~[vertx-core-3.3.0.jar:na]
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:339) ~[netty-common-4.1.1.Final.jar:4.1.1.Final]
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:393) ~[netty-transport-4.1.1.Final.jar:4.1.1.Final]
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:742) ~[netty-common-4.1.1.Final.jar:4.1.1.Final]
	at java.lang.Thread.run(Unknown Source) ~[na:1.8.0_91]

_Copied from original issue: eclipse/vert.x#1482_

pmlopes:
…the pipeline will get out of order and responses are not aligned with handlers

The client was using a concurrent list to safeguard from concurrent access to the handler/response queue, however the proper control should be around the queue and the socket writes a write must be properly aligned with a offer to the queue otherwise the waiting handlers will be offset (the cast exception).

### DIFF
--- a/src/main/java/io/vertx/redis/impl/RedisConnection.java
+++ b/src/main/java/io/vertx/redis/impl/RedisConnection.java
@@ -248,10 +248,14 @@ class RedisConnection {
         break;
       case CONNECTING:
       case ERROR:
-        pending.add(command);
+        runOnContext(v -> {
+          pending.add(command);
+        });
         break;
       case DISCONNECTED:
-        pending.add(command);
+        runOnContext(v -> {
+          pending.add(command);
+        });
         connect();
         break;
     }
--- a/src/main/java/io/vertx/redis/impl/RedisConnection.java
+++ b/src/main/java/io/vertx/redis/impl/RedisConnection.java
@@ -137,23 +137,21 @@ class RedisConnection {
       replyParser.reset();
 
       client.connect(config.getPort(), config.getHost(), asyncResult -> {
-        //context = vertx.getOrCreateContext();
-
         if (asyncResult.failed()) {
           runOnContext(v -> {
-            state.set(State.ERROR);
+            if (state.compareAndSet(State.CONNECTING, State.ERROR)) {
+              // clean up any waiting command
+              clearQueue(waiting, asyncResult.cause());
+              // clean up any pending command
+              clearQueue(pending, asyncResult.cause());
 
-            // clean up any waiting command
-            clearQueue(waiting, asyncResult.cause());
-            // clean up any pending command
-            clearQueue(pending, asyncResult.cause());
+              // close the socket if previously connected
+              if (netSocket != null) {
+                netSocket.close();
+              }
 
-            // close the socket if previously connected
-            if (netSocket != null) {
-              netSocket.close();
+              state.set(State.DISCONNECTED);
             }
-
-            state.set(State.DISCONNECTED);
           });
         } else {
           netSocket = asyncResult.result()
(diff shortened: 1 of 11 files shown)


pmlopes:
TODO: synchronize the state changes

vietj:
the `pending` queue can be accessed from two different threads in `resetPending` method and `send` method.

src/main/java/io/vertx/redis/impl/RedisConnection.java
```diff
@@ -248,10 +248,14 @@ void send(final Command<?> command) {
         break;
       case CONNECTING:
       case ERROR:
-        pending.add(command);
+        runOnContext(v -> {
+          pending.add(command);
```
vietj: it should actually be a real synchronization without `runOnContext` because you can run in the case with `send` can be called two times: the `pending` list is processed while having context tasks that will add command to the pending queue.

and this should be done along with the `state` field, i.e you append to `pending` only if the state is CONNECTING or ERROR and at connection time, the state is updated to CONNECTED and the pending queue is emptied atomically (which means we should not use an AtomicReference.

pmlopes: I guess that we could avoid the `synchronized` at all if we would double check the state.

All reads/writes to the queue happen inside a `runOnContext` block, which means we're always in the same thread, it could have happen that the state changed in between the start of the `send` method and the `runOnContext`, so if on that case I'd do a double check such as:

```java
runOnContext(v -> {
  if (state.get() != CONNECTED) {
    pending.add(command);
  } else {
    // state changed so start over...
    send(command);
  }
});
```

vietj: yes that's a possibility, however I'm wondering if we could not end up with commands executed in the wrong order...

vietj: my point is that such design create sometimes complex state machines and we can have subtle bugs whereas just using synchronized makes it easier to reason about

PR #76, opened 2017-03-28
Fixes #69

### LINKED ISSUE #69: can't continue subscribe while reconnected
### Version

* vert.x core:3.3.3
* vert.x redis client:3.3.3

### Context

can't continue subscribe while reconnected, or how to handle connection's reconnect event?

pmlopes:


src/main/java/io/vertx/redis/impl/RedisConnection.java
```diff
@@ -144,10 +143,17 @@ else if (data.length == 4) {
   }
 
   private void connect() {
+    // in case the user has disconnected before, update the state
+    reconnect.compareAndSet(false, true);
```
vietj: use `reconnect.set(true)` instead

src/main/java/io/vertx/redis/impl/RedisConnection.java
```diff
@@ -84,14 +85,15 @@
   }
 
   private final AtomicReference<State> state = new AtomicReference<>(State.DISCONNECTED);
+  // attempt to reconnect on error, by default true
+  private final AtomicBoolean reconnect = new AtomicBoolean(true);
```
vietj: it seems there is no real need for atomic, use `volatile` rather

src/main/java/io/vertx/redis/RedisOptions.java
```diff
@@ -18,169 +18,209 @@
...
- *
+ * <p>
  * The usage of this two extra properties is to setup required authentication and optionally the selection of the active
  * database at connection time. If you define this extra properties on every connection to Redis server this client
  * will perform the authentication handshake and database selection, however if you don't do this and call {@link io.vertx.redis.RedisClient#auth(String, Handler)}
  * yourself in case of connection failure the client will not be able to perform the correct authentication handshake.
  */
-@DataObject
-public class RedisOptions {
+@DataObject(generateConverter = true)
```
vietj: it should be "no" as the NetClientOptions handle the json conversion itself

src/main/java/io/vertx/redis/RedisOptions.java
```diff
@@ -18,169 +18,209 @@
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.core.Handler;
 import io.vertx.core.json.JsonObject;
+import io.vertx.core.net.NetClientOptions;
 
 /**
  * This object controls the connection setting to the Redis Server. There is no need to specify most of the settings
  * since it has built the following sensible defaults:
- *
+ * <p>
```
vietj: actually this is fine to have HTML in asciidoc

PR #78, opened 2017-03-28
Sync API with 3.2.1

pmlopes:


src/main/java/io/vertx/redis/impl/RedisClientImpl.java
```diff
@@ -34,7 +34,7 @@
 
 public final class RedisClientImpl extends AbstractRedisClient {
 
-  RedisTransaction transaction;
+  private RedisTransaction transaction;
```
cescoffier: Can be considered as a breaking change. However, I think it's fine in this case.

pmlopes: it is in the `impl` package, but can be reverted without issues

pmlopes:
Unless stated on the redis.io docs itself all new APIs are backward compatible, i mean, attempting to call a new command on an old redis instance will end up with the response `ERR description` which in our case will be handled and a `AsyncResult` failure with that message.

PR #79, opened 2017-04-29
Add high level scripting method

szdavid92:
Adds the method `evalScript` to `RedisClient`. 

This is not a Redis command, rather a higher level function that can be used to evaluate a script on Redis without worrying about script caching. It is modelled after [defineCommand in luin/ioredis](https://github.com/luin/ioredis#lua-scripting), and basically serves as a replacement for `eval`.
When calling `evalScript` it executes EVALSHA, and if the server responds with NOSCRIPT error, it tries again using EVAL. (This makes it impossible to do in a transaction).

See [Bandwith and EVALSHA](https://redis.io/commands/eval#bandwidth-and-evalsha)

### DIFF
--- a/src/main/java/examples/Examples.java
+++ b/src/main/java/examples/Examples.java
@@ -5,6 +5,7 @@ import io.vertx.core.json.JsonObject;
 import io.vertx.redis.RedisClient;
 import io.vertx.redis.RedisOptions;
 import io.vertx.redis.RedisTransaction;
+import io.vertx.redis.Script;
 
 /**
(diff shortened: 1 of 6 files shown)


vietj:
@pmlopes what do you think ?

pmlopes:
@szdavid92 could you please rebase your PR to the current master? after that I think we can merge

szdavid92:
@pmlopes done

pmlopes:
Thanks!

PR #84, opened 2017-08-04
fix API typpo, set binary should take Buffer

### LINKED ISSUE #83: Expand binary support for Redis
### Version

* vert.x core: 3.4.2
* vert.x redis client: 3.4.2

### Context
We are using Redis to store constantly incoming binary data, storing it to file is not an option since there might be race conditions and information will be lost. The only append function Redis currently have is for String. Since you have setBinary function I figured I can get the current data in form of Buffer using getBinary function, append to the buffer and use setBinary, but to avoid race condition I must use transaction.multi, and thats where the biggest issue is, ## getBinary in RedisClient class returns a Buffer, but getBinary in RedisTransaction class returns a String. Converting byte [] to String ruins the data so its not an option.

pmlopes:
Fixes #83

### DIFF
--- a/src/main/java/io/vertx/redis/RedisTransaction.java
+++ b/src/main/java/io/vertx/redis/RedisTransaction.java
@@ -813,7 +813,7 @@ public interface RedisTransaction {
    * group: string
    */
   @Fluent
-  RedisTransaction getBinary(String key, Handler<AsyncResult<String>> handler);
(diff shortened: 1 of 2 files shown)


tsegismont:
How about adding a test?

vietj:
it should be documented as minor _breaking change_

pmlopes:
This is just sends the request as is to the redis transaction queue, The type change is handled by the generic parser and already covered by the non transaction tests

Issue #90, opened 2017-10-12
Support for Unix sockets


ankushs92:
Hey guys,
We've been using vertx-redis-client for a while in production now, and so far everything is great. 
We are currently exploring libraries that provide support for connecting to Redis via Unix sockets. Any plans on providing this feature in the near future?

vietj:
@ankushs92 Vert.x 3.5 provides support for Unix domain sockets and the actual Redis client uses a NetClient, however it does not expose the domain socket setting but it should be easy to achieve, would you mind to contribute such feature ?

BillyYccc:
I made a PR to enable the domain socket option in Redis client.

pmlopes:
Fixed by #91

Issue #68, opened 2016-10-07
make client less opionated


pmlopes:
Currently the client will behind the scenes enqueue request until it manages to connect to the server, it will try to reconnect on error and hold messages that have not been processed and attempt to handle them once a connection is available. It will also switch from pub/sub mode to normal mode and establish 2 connections so regular calls do not interfere with each other.

Although this is quite nice for people who just want it to work, for the case where one wants total control of what is going on this is bad decision.

I'd propose to create a different client class that is just purely simple. Establish a connection and has several handlers, for exception, error, etc, and sends commands.

jmaasing:
It is also not currently possible to set Net client options such as idle timeout. Meaning if redis hangs the client hangs. It would be useful if the user of the API could have control over the connection settings.

vietj:
so you mean something like a `RedisConnection` ?

pmlopes:
Yes. Full control over the connection with handlers for connect, disconnect, exception plus using the core options for socket creation. And the only extra method would be one that could send a arbitrary command to redis and parse the response.

vietj:
I would be in favor of this, see my other comment on the other issue (about having an SQLOperations like pattern for both connection/client).

PR #91, opened 2017-10-15
Add support for domain socket

BillyYccc:
Signed-off-by: BillyYccc <billy112487983@gmail.com>

This work is for Issue #90 , make it possible to use domain socket in Redis client.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,12 +24,12 @@
   </parent>
 
   <artifactId>vertx-redis-client</artifactId>
-  <version>3.5.0-SNAPSHOT</version>
+  <version>3.5.1-SNAPSHOT</version>
 
   <name>Vert.x Redis Client</name>
 
   <properties>
-    <stack.version>3.5.0-SNAPSHOT</stack.version>
+    <stack.version>3.5.1-SNAPSHOT</stack.version>
     <codegen.rxjava.deprecated>true</codegen.rxjava.deprecated>
   </properties>
 
--- a/src/main/asciidoc/dataobjects.adoc
+++ b/src/main/asciidoc/dataobjects.adoc
@@ -214,12 +214,20 @@ Set the password for authentication at connection time.
 +++
 |[[binary]]`binary`|`Boolean`|
 +++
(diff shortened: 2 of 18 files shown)


vietj:
have you tested it ?

BillyYccc:
@vietj Yeah, I have tested this in Ubuntu 16.04, it seems to works fine with Redis 4.0.2.

vietj:
ok we will merge the PR after 3.5.0 release, as for now it's too late to have it, it shall be in 3.5.1.

the concern I have is the declaration in options of domain socket, currently we don't provide a way to have a default socket address for domain sockets and I believe it is fine because one can use `NetClient#connect(SocketAddress)`, so if the `RedisClient` has a similar connect method it would be enough I think.

BillyYccc:
The vertx-redis-client at present uses a mechanism of automatic reconnection, the client connects to Redis server as soon as it is created with a config. Therefore I think if we use a new connect method like `RedisClient#connect(SocketAddress)`, this might bring breaking changes to the users. WDYT?

vietj:
I think that simply overloading 

```
static RedisClient create(Vertx vertx, RedisOptions config, SocketAddress address)
```

would be enough overall

pmlopes:
It’s a bigger discussion than just this issue but I think that in our APIs where we have ‘options, otherType’ in the create methods it feels like a code smell since the options are not a good data structure to define all options as we still need to depend on other data types and variables to configure the way we want the object to behave.

pmlopes:
I think the issue #68 needs to be prioritized as it starts to make more sense. API independence and full control over the connection, no lazy connect, no auto reconnect. The user would have control over this behavior (which is not available) currently.

vietj:
@pmlopes I agree with you, that it's good to provide a simple API like RedisClient and also something that provides more control like a `RedisConnection`

Overall we could adopt the same pattern we introduced for `SQLClient` and have a `RedisOperations` interface that is both on `RedisConnection` and `RedisClient`.

BillyYccc:
Thanks for reviewing!

PR #101, opened 2018-04-10
Make a less opionated Redis client

### LINKED ISSUE #68: make client less opionated

pmlopes:
Fixes #68 

See: https://www.jetdrone.xyz/2018/08/06/Vertx-Redis-API.html

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -24,12 +24,12 @@
   </parent>
 
   <artifactId>vertx-redis-client</artifactId>
-  <version>3.6.0</version>
+  <version>4.0.0-SNAPSHOT</version>
 
   <name>Vert.x Redis Client</name>
 
   <properties>
-    <stack.version>3.6.0</stack.version>
+    <stack.version>4.0.0-SNAPSHOT</stack.version>
   </properties>
 
   <!--


src/main/java/io/vertx/redis/Args.java
```diff
@@ -0,0 +1,132 @@
...
+  @Fluent
+  Args addNull();
+
+  /**
+   * Adds a byte array
+   * @return self
+   */
+  @GenIgnore
+  @Fluent
+  default Args add(byte[] arg) {
```
vietj: I think we should have a `Buffer` version as well.

src/main/java/io/vertx/redis/Args.java
```diff
@@ -0,0 +1,132 @@
...
+    return add(Buffer.buffer(arg));
+  }
+
+  /**
+   * Adds a String argument
+   * @return self
+   */
+  @Fluent
+  default Args add(String arg) {
+    if (arg == null) {
```
vietj: `null` should be considered as a programming error instead

src/main/java/io/vertx/redis/Reply.java
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * If the reply is a not a multibulk it return the original buffer.
+   * @return buffer
+   */
+  @Nullable
+  Buffer asBuffer();
+
+  /**
+   * Is this reply a multu bulk?
```
vietj: multu -> multi

src/main/java/io/vertx/redis/Args.java
pmlopes: The reason i choose it this way was because if you have something like:
```java
String s = null; // or as a result of a operation that returns null
args.add(s); // should be ok
```

vietj:
it 's not clear what this PR does

- does it implement the parts we discussed in roadmap ?
- is it a complement or replacement to the existing API ?

pmlopes:
This PR provides what we've discussed during the roadmap: *A non opinionated redis client*

1. The connection is not auto-magically managed by the client
2. The client explicitly requires users to connect, disconnect
3. Allows handlers for `close` and `exceptions`
4. No API mappings to redis commands (meaning no need to maintain on every redis release)
5. No hard coded types which allow usage of buffers (this gives better performance and more flexibility, for example when implementing a web session cache we can rely on the current session encoding (Buffer) instead of having to force a custom codec that writes a string.

Having this as a core client will allow further evolution (not necessarily in this project) to support, HA, clustering, etc... or even to implement other things.

vietj:
I'm fine to keep a single client at the moment.

it would be nice to have the old one reimoplemneted on this of the new one as it makes more testing and finding bugs and also less similar code to maintain.

the question is : what do we want to do with the actual version on the long term ?

PR #113, opened 2018-07-24
Return null when getting a key that does not exist

okou19900722:
https://github.com/vert-x3/vertx-lang-kotlin/issues/69


pmlopes:
Thanks!

PR #123, opened 2018-12-17
squashed all commits

pmlopes:
This is the follow up from: https://www.jetdrone.xyz/2018/08/06/Vertx-Redis-API.html

In this PR we depart from the unmaintainable manual crafted API to a generic API, with focus on simplicity. The old API is still present and marked as deprecated and using the new implementation under the hood.


vietj:
not yet on the full review however I see that Command in an enum, since recently it is possible to have constant in interface in `@VertxGen`, using such scheme I think would make it easier to add external commands if we use something like:

```
@VertxGen
public interface Command {
   Command APPEND = Command.create(...);
}
```

pmlopes:
@vietj agreed. I've converted the `enum` to an `interface` which also allows making the API future proof, if new commands are added, users do not need to wait for new vertx releases, they can just fill in the metadata.

src/main/java/io/vertx/redis/client/RedisException.java
```diff
@@ -0,0 +1,88 @@
...
+ */
+package io.vertx.redis.client;
+
+/**
+ * Maps a redis error message to a JVM Exception type.
+ * Redis Errors have no trace so this throwable will not contain a stacktrace.
+ *
+ * @author Paulo Lopes
+ */
+public final class RedisException extends Throwable {
```
vietj: why extends Throwable ? you can get the no stack trace property with regular exceptions too

src/main/java/io/vertx/redis/client/RedisOptions.java
```diff
@@ -0,0 +1,187 @@
...
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.json.JsonObject;
+import io.vertx.core.net.NetClientOptions;
+import io.vertx.core.net.SocketAddress;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@DataObject(generateConverter = true)
+public class RedisOptions {
```
vietj: this class is missing javadoc on the class and its members

src/main/java/io/vertx/redis/client/RedisRole.java
```diff
@@ -13,10 +13,13 @@
...
+package io.vertx.redis.client;
 
-/**
- * @author <a href="mailto:marko.strukelj@gmail.com">Marko Strukelj</a>
- */
-public interface RedisPost28Tests {
+import io.vertx.codegen.annotations.VertxGen;
+
+@VertxGen
+public enum RedisRole {
```
vietj: missing javadoc

src/main/java/io/vertx/redis/client/RedisSlaves.java
```diff
@@ -13,11 +13,13 @@
...
 
-public enum ResponseTransform {
-  NONE,
-  HASH,
-  INFO,
-  ARRAY
+import io.vertx.codegen.annotations.VertxGen;
+
+@VertxGen
+public enum RedisSlaves {
```
vietj: missing javadoc

src/main/java/io/vertx/redis/client/Request.java
```diff
@@ -0,0 +1,145 @@
...
+ * </blockquote>
+ *
+ * So all non String/Bulk types will be encoded to Bulk for convenience.
+ *
+ * @author Paulo Lopes
+ */
+@VertxGen
+public interface Request {
+
+  static Request cmd(Command commandEnum) {
```
vietj: it's not anymore an enum

src/main/java/io/vertx/redis/client/Request.java
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  /**
+   * Adds a NULL encoded string
+   * @return self
+   */
+  @Fluent
+  Request nullArg();
+
+  Command command();
```
vietj: missing javadoc

src/main/java/io/vertx/redis/client/Response.java
```diff
@@ -0,0 +1,109 @@
...
+import io.vertx.codegen.annotations.GenIgnore;
+import io.vertx.codegen.annotations.VertxGen;
+import io.vertx.core.buffer.Buffer;
+
+import java.nio.charset.Charset;
+import java.util.Iterator;
+import java.util.Set;
+
+@VertxGen
+public interface Response extends Iterable<Response> {
```
vietj: missing javadoc, also the javadoc should explain why a response is an iterable of response (recursion ?)

src/main/java/io/vertx/redis/client/ResponseType.java
```diff
@@ -13,12 +13,15 @@
...
-/**
- * An interface mainly created to pass 2 parameters to pub/sub message handler (thus avoiding double casts)
- */
-@FunctionalInterface
-public interface MessageHandler {
-  void handle(String channelOrPattern, Reply[] replyData);
+import io.vertx.codegen.annotations.VertxGen;
+
+@VertxGen
+public enum ResponseType {
```
vietj: javadoc

vietj:
the code looks great, however it is missing documentation:

- javadocs
- I see many deprecated classes, the deprecated should use the `@deprecated` to indicate what should be used instead

PR #127, opened 2019-01-21
Issues/68 Adding support for clusterwide commands

pendula95:
Added support for clusterWide commands, example command KEYS. This commands should be executed on all nodes (nodes that contain different slot).
Also need to think how to implement this feature for commands that need to be executed on all nodes (not only different slot nodes). Example command SAVE.


pendula95:
Have you had a chance to review this? I see it has conflicts now, to which branch do you want me update this?

pendula95:
There is a difference, for example `keyless` command CLUSTER NODES can be executed on any node and it will give you same result regardless of the node it has been executed on so yes you can execute it on random node. But for example command SCAN gives different results depending on which node it has been executed or command FLUSHDB (which is keyless). For these commands to have any meaning they need to executed on all nodes and than result can be reduced. 

By having the reducers that tells us if command should be executed on all nodes we have nearly same effect as editing the command list as we have to maintain that list manually but yes in some way it is decoupled from the auto generated command list. If you are ok we can implement logic if REDUCER PRESENT && COMMAND IS KEYLESS execute on all nodes.

pmlopes:
I believe this is the best we should do. Keeping the current command list, allows us to do code generation and API sync in one click. Implementing the `isKeyless() && hasReducer()` to run on all nodes then allows fixing this issue and keep the code decoupled.

Let's go this way! :+1:

PR #129, opened 2019-03-14
Use a single factory

pmlopes:
This will allow instances of redis clients to be configurable instead of hard coded.

### DIFF
--- a/src/main/java/io/vertx/redis/client/Redis.java
+++ b/src/main/java/io/vertx/redis/client/Redis.java
@@ -65,35 +65,6 @@ public interface Redis extends ReadStream<Response> {
   }
 
   /**
-   * Connect to redis in sentinel mode, the {@code onConnect} will get the {@link Redis} instance.
-   *
-   * This is a helper factory which uses a preset of defaults for the connection:
-   *
-   * <ul>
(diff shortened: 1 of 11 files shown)


src/main/java/io/vertx/redis/client/Redis.java
```diff
@@ -46,35 +49,48 @@ static void createClient(Vertx vertx, SocketAddress address, Handler<AsyncResult
...
+   * Connect to redis in cluster mode, the {@code onConnect} will get the {@link Redis} instance.
+   *
+   * This is a helper factory which uses a preset of defaults for the connection:
+   *
+   * <ul>
+   *     <li>client type - <b>CLUSTER</b></li>
+   *     <li>slave usage - <b>NEVER</b></li>
+   * </ul>
    */
   static void createClusterClient(Vertx vertx, SocketAddress address, Handler<AsyncResult<Redis>> onCreate) {
```
tsegismont: Do we need to keep these methods since we have all what's needed in the base method?

pmlopes: I was just keeping them as shortcuts, but they can go.

Issue #132, opened 2019-03-19
Lazy Redis command queuing


vietj:
The new redis client API is not lazy anymore and it might create usability issues when it comes to queue pending request when a service handle multiple requests.

When a service (Verticle) will create a Redis instance it will need to queue commands until the client is connected (it could be done initially but that would not work if the TCP connection is closed abruptly).

So the service will have to manage:

- concurrent connection to a redis instance
- queue redis commands until the connection succeeds or fails
- dequeue redis commands when the asynchronous connection result happens

We need to make this possible in the client itself instead of putting the burden on developers.

vietj:
for the record here is what is required in vertx-config-store to make it work like before https://github.com/vert-x3/vertx-config/commit/aa3db574e7b3ea31c2325d040c4145bfc9a1c5fa

vietj:
possible ways to achieve this:

- `createShared` like function that would handle this, where the user would then close the Redis after use to release it
- make a pool of redis connection (a proxy) of 1 connection (which is similar to the `createShared` option but is not static)

KowalczykBartek:
hi @vietj - I want to play around that but just have a question - is RedisClient designed to be used only from the single thread ?
I didn't found nothing about that in docu, but looking at code seems to be not designed to be thread safe
```java
  @Override
  public Redis send(final Request request, Handler<AsyncResult<Response>> handler) {
    //stuff ....
    try {
      // encode the message to a buffer
      final Buffer message = ((RequestImpl) request).encode();
      // write to the socket
      waiting.offer(handler); /// <------ not atomic
      netSocket.write(message); ;// <------ not atomic
    } catch (RuntimeException e) {
      onException.handle(e);
    }
    return this;
  }
```
and
```java
  // waiting: commands that have been sent but not answered
  // the queue is only accessed from the event loop
  private final ArrayQueue waiting; // not thread safe also 
```

KowalczykBartek:
Hi - 
What do you think about such approach https://github.com/KowalczykBartek/vertx-redis-client/pull/1 ?
I tried to integrate it with redis-config and it looks that it will work :) 

Please give me some feedback - so I can proceeded with the rest and with tests.

BTW: 
```
    final AtomicInteger count = new AtomicInteger(commands.size());
    final AtomicBoolean failed = new AtomicBoolean(false);
```
I think that this can be replaced with something that do not enforce write/read barrier - especially that this is always executed from event loop.

Thanks !

[...]

PR #135, opened 2019-03-19
Remove deprecation for now in 3.7

vietj:
this PR removes RedisClient deprecation in 3.7 as it seems premature because the new API lacks of usability when it comes to concurrently using the new Redis API putting all the burden on the user to manage concurrent connection to redis.

We need to make another iteration on the new redis API to provide the same usability before deprecating the redis client.


vietj:
See https://github.com/vert-x3/vertx-redis-client/issues/132

vietj:
ping @pmlopes

PR #145, opened 2019-04-12
All slot commands

pendula95:
Closes #127


pmlopes:
The tests fail, can you check what is going on? I belive that after that we can merge

PR #156, opened 2019-05-17
initial import

pmlopes:


vietj:
I would like this code to use the existing pool that is in vertx-core `io.vertx.core.http.impl.pool.Pool`, although it is named HTTP I believe it is generic enough to be used for any kind of non-blocking pool.

vietj:
we shall eventually rename the pool name in core to not be in the HTTP package too of course.

pmlopes:
I'll solve the conflicts and replace the pool with the http pool

pmlopes:
@vietj I looked at the pool, there are 2 differences, the http pool doesn't offer a shared mode (across verticles, like we do for jdbc, cassandra, mongo...) what are the ideas here, are we supposed to avoid having shared pools?

/cc @tsegismont

vietj:
yes we should try avoid sharing pools.

we don't have any pool yet with shared mode, so we need to think of a way to share pools between verticles for now try to make it not shared and we will brainstorm on shared pools separately.

vietj:
that being said I think that an initial implementation would be to have a shared pool to use a specific event loop and then connections borrowed from this shared pool will simply do message passing between the client and the event loop it belongs to.

taht's what the new SQL client is doing.

pmlopes:
This PR has been replaced by: https://github.com/vert-x3/vertx-redis-client/pull/173

PR #161, opened 2019-05-29
fixes #160 Cluster Client send READONLY command on connection proposal

### LINKED ISSUE #160: Redis Cluster avoid move response on ALWAYS or SHARE slave mode
### Version

* vert.x core: 3.7.0
* vert.x redis client: 3.7.0

### Context
When the client is configured in cluster mode and the slave mode is ALWAYS or SHARE
Ex.
`RedisOptions options = new RedisOptions()
    .setType(RedisClientType.CLUSTER)
    .setUseSlave(RedisSlaves.SHARE)`

The client when calls a slave always receive MOVED response from Redis and client does a redirect on a Master node. To avoid this behavior the client should be call READONLY command on slaves node at connection time as seen on Redis documentation [READONLY Command](https://redis.io/commands/readonly)

joker-90:
This is PR related to issue #160

Following discussion on Gitter between @damianofontana85 and @pmlopes I have implemented a possible solution for the related issue. The redis cluster client sends READONLY command if the slave mode is ALWAYS or SHARE to slave nodes at connection time.

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
@@ -619,8 +619,7 @@ public class RedisClusterClient implements Redis {
     }
   }
 
-  private void send(final Redis client, final RedisOptions options, final int retries, Request
-    command, Handler<AsyncResult<Response>> handler) {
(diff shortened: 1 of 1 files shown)


sreejit123:
@pmlopes can this fix be available in 3.7.0? We have some issues regarding overall vertx upgrade and switching to 3.8.0 involves some breaking changes

PR #163, opened 2019-06-18
IntegerType: override Response.toInteger()

### LINKED ISSUE #162: Response.toInteger() always requires round-tripping through String
Because `IntegerType` does not override `Response.toInteger()`, the default implementation applies:

https://github.com/vert-x3/vertx-redis-client/blob/master/src/main/java/io/vertx/redis/client/Response.java#L74-L80

This will wastefully convert the value to a string and parse it back as an integer when it could just be returned directly. A truncating cast would have to be applied, but in cases where the value is not expected to be out of range this is fine; we could even assert that the value is within the range of `int` in debug mode as an anti-footgun.

abonander:
closes #162

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/types/IntegerType.java
+++ b/src/main/java/io/vertx/redis/client/impl/types/IntegerType.java
@@ -34,6 +34,11 @@ public final class IntegerType implements Response {
   public ResponseType type() {
     return ResponseType.INTEGER;
   }
+  
+  @Override
+  public Integer toInteger() {
+    return value.intValue();
+  }
 
   @Override
   public Long toLong() {


src/main/java/io/vertx/redis/client/impl/types/IntegerType.java
```diff
@@ -34,6 +34,11 @@ private IntegerType(long value) {
   public ResponseType type() {
     return ResponseType.INTEGER;
   }
+  
+  @Override
+  public Integer toInteger() {
+    return (int) value;
```
abonander: RFC: do we want to report somehow when the value is out of range for `int`?

* use `assert` so it only throws when assertions are enabled
* throw `IllegalStateException`
* log an error/warning

Any of these checks would impose overhead but not nearly as much as round-tripping through a string.

Some users may also intend to perform a truncating cast so maybe we should only return the value.

src/main/java/io/vertx/redis/client/impl/types/IntegerType.java
```diff
@@ -34,6 +34,11 @@ private IntegerType(long value) {
   public ResponseType type() {
     return ResponseType.INTEGER;
   }
+  
+  @Override
+  public Integer toInteger() {
+    return value.intValue();
```
abonander: RFC: do we want to report somehow when the value is out of range for `int`?

* use `assert` so it only throws when assertions are enabled
* throw `IllegalStateException`
* log an error/warning

Any of these checks would impose overhead but not nearly as much as round-tripping through a string.

Some users may also intend to perform a truncating cast so maybe we should only return the value.

abonander:
@pmlopes that was my only question so this is good to merge now.

abonander:
@pmlopes I do have a question about the API design however; was it a deliberate decision to use boxed types instead of primitives (`Long` vs `long`, etc.)?

pmlopes:
Thanks! :+1:

PR #164, opened 2019-06-19
Fix null pointer onException.handle call

aygalinc:
Replace direct call by fail method call

Signed-off-by: aygalinc <colin.aygalinc@gmail.com>


pmlopes:
Thanks!

PR #165, opened 2019-06-24
Futurization

pmlopes:
Upgrade API to support vert.x 4.0 Future


vietj:
are there APIs that needs to be deprecated here ?

tools/redis-api.hbs
```diff
@@ -37,13 +37,25 @@ public interface RedisAPI {
...
+  /**
+   * Redis command <a href="https://redis.io/commands/{{ name }}">{{ camelcase safename }}</a>.
+   * @return Future response.
+   */
+  default Future<@Nullable Response> {{ camelcase safename }}({{{ types }}}) {
+    return send(Command.{{ enum }}{{#if argLen}}, {{{ args }}}{{#if variable}}.toArray(){{/if}}{{/if}});
+  }
 {{/each}}
 
   @GenIgnore
```
vietj: should't we remove @GenIgnore ?

pmlopes: we can't this a `varargs` method which is handy to delegate all synthetic API calls as the compiler enforces the types

vietj: oh right, we need to support varargs at some point in codegen :-)

PR #170, opened 2019-07-02
Fix cluster batching for single key commands (#169)

### LINKED ISSUE #169: Batching single key commands fails in cluster mode
### Version

* vert.x core: 3.7.1
* vert.x redis client: 3.7.1

### Context

When running a batch of redis commands with single key, the execution fails:

```
io.vertx.core.impl.NoStackTraceThrowable: RedisClusterClient does not handle batching commands with keys across different slots. TODO: Split the command into slots and then batch.
```
Batched commands are:

```
MULTI
INCR <key>
EXPIRE <key> <seconds>
EXEC
```

### Do you have a reproducer?

Can provide one if needed.

### Steps to reproduce

1. Execute a batch of single key redis commands

### Extra

The problem is in `RedisClusterClient` line 387.

image

If the command has a single key, it is checked against the current slot - which is `-1` initially.
The separate guard condition for `currentSlot == -1` is missing.

I think I can workaround by copying/extending the redis client and using the patched versions.
If there is a chance to get a fix into a 3.x.x release, I'd be happy to provide a PR.

pyranja:
Fixes #169

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
@@ -382,7 +382,13 @@ public class RedisClusterClient implements Redis {
 
       // last option the command is single key
       int start = cmd.getFirstKey() - 1;
-      if (currentSlot != ZModem.generate(args.get(start))) {
+      final int slot = ZModem.generate(args.get(start));
+      // we are checking the first request key
+      if (currentSlot == -1) {
+        currentSlot = slot;
+        continue;
+      }
+      if (currentSlot != slot) {
         // in cluster mode we currently do not handle batching commands which keys are not on the same slot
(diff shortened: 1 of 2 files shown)


pyranja:
Is there a guide to set up a dev environment, I'd like to add a test for this, but I'm missing the vertx snapshot stack.

pmlopes:
I've this on my `~/.m2/settings.xml`:

```xml
<settings>
  <profiles>
    <profile>
      <id>allow-snapshots</id>
      <activation><activeByDefault>true</activeByDefault></activation>
      <repositories>
        <repository>
          <id>snapshots-repo</id>
          <url>https://oss.sonatype.org/content/repositories/snapshots</url>
          <releases><enabled>false</enabled></releases>
          <snapshots><enabled>true</enabled></snapshots>
       </repository>
     </repositories>
   </profile>
  </profiles>
</settings>
```

This should give you snapshots for most OSS projects

pyranja:
Thanks, I got it working with the sonatype repository.

I added tests to cover the modified logic, please let me know if there is sth. to improve.

PR #173, opened 2019-07-05
New pool

pmlopes:
This is a new pool based implementation using the core http pool

This is a follow up from: https://github.com/vert-x3/vertx-redis-client/pull/156

A new PR was started as the master branch and the original PR drifted to much and it was decided to use a different kind of base pool.


pmlopes:
@vietj I need you to review specially: https://github.com/vert-x3/vertx-redis-client/pull/173/files#diff-1a75e8ea2e28b9f3cced2c1c431e5b14

As this pool was new to me, I'd like to be sure that I'm using it the correct way...

vietj:
sure, I'll move the current HTTP pool package to a more generic package (like io.vertx.core.impl) at some point too.

Julien

src/main/java/io/vertx/redis/client/RedisConnection.java
```diff
@@ -0,0 +1,146 @@
...
+import io.vertx.core.*;
+import io.vertx.core.streams.ReadStream;
+
+import java.util.List;
+
+/**
+ * A simple Redis client.
+ */
+@VertxGen
+public interface RedisConnection extends ReadStream<Response> {
```
vietj: I think this class should not copy the javadoc from ReadStream and instead inherit the doc from the parent interface.

also it should use `@Override` annotation on methods

src/main/java/io/vertx/redis/client/impl/RedisConnectionImpl.java
```diff
@@ -0,0 +1,308 @@
...
+  }
+
+  void forceClose() {
+    netSocket.close();
+  }
+
+  @Override
+  public void close() {
+    // recycle this connection from the pool
+    listener.onRecycle(15_000L);
```
vietj: make this configurable.

PR #190, opened 2019-09-23
Fixes #189

### LINKED ISSUE #189: New connection pool seems broken
### Version

* vert.x core:4.0.0-milestone3
* vert.x redis client: 4.0.0-milestone3

### Context

I made a simple verticle that every 1s takes and releases a connection. After 4 executions, acquiring connection fails with io.vertx.core.http.ConnectionPoolTooBusyException: Connection pool reached max wait queue size of 1

### Do you have a reproducer?
```java

public class RedisVerticle extends AbstractVerticle {
    @Override
    public void start() {
        Redis client = Redis.createClient(vertx);
        vertx.setPeriodic(1000, event -> {
            client.connect(event1 -> {
                if (event1.succeeded()) {
                    System.out.println("Connection acquired");
                    event1.result().close();
                } else
                    event1.cause().printStackTrace();
            });
        });
    }
}
```

> Connection acquired
Connection acquired
Connection acquired
Connection acquired
io.vertx.core.http.ConnectionPoolTooBusyException: Connection pool reached max wait queue size of 1
	at io.vertx.core.http.impl.pool.Pool.lambda$nextTask$1(Pool.java:300)
	at io.vertx.core.http.impl.pool.Pool.checkPendingTasks(Pool.java:256)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:163)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:416)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:515)
	at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:918)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.base/java.lang.Thread.run(Thread.java:834)

zenios:
1.Recycle timeout is relative to currentTimeMillis
2.Actually close the connection when asked from pool

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/ConnectionManager.java
+++ b/src/main/java/io/vertx/redis/client/impl/ConnectionManager.java
@@ -167,6 +167,7 @@ class ConnectionManager {
       connection.handler(null);
       connection.endHandler(null);
       connection.exceptionHandler(null);
+      ((RedisConnectionImpl) connection).forceClose();
     }
   }
 
--- a/src/main/java/io/vertx/redis/client/impl/RedisConnectionImpl.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisConnectionImpl.java
@@ -50,7 +50,8 @@ public class RedisConnectionImpl implements RedisConnection, ParserHandler {
   @Override
   public void close() {
     // recycle this connection from the pool
-    listener.onRecycle(recycleTimeout);
+    long expired = recycleTimeout > 0 ? System.currentTimeMillis() + recycleTimeout : 0L;
+    listener.onRecycle(expired);
   }
 
   @Override


pmlopes:
:+1: for the PR

Perhaps you could add the test case too to verify the change works and we won't get a regression.

zenios:
I am not sure whether adding tests for those two line will do any difference. 

Both lines are part of the requirements in order for the pool to properly work (Pool is already tested in vertx-core)

Most probably what is missing are tests for the whole ConnectionManager which in order to be properly tested, it has to change from protected to public and also a method to expose the actual pool out of the manager in order to fetch metrics and verify while testing

Please correct me if i am wrong. Testing is not my strong point

pmlopes:
Thanks!

Issue #188, opened 2019-09-22
RedisClient no longer publishes received pub/sub messages to event bus


guss77:
### Version

* vert.x core: 3.8.1
* vert.x redis client: 3.8.1

### Context

Up until Vert.x Redis Client 3.6.3, users using the `RedisClient` class (that implements auto-reconnect and other useful features) could call the `client.subscribe(channel)` method, after which clients can listen to the Vert.x event bus to receive published messages.

Starting with Vert.x Redis Client 3.7, this functionality isn't working anymore - messages received on subscribed pub/sub channels are not forwarded to the event bus, instead, we see this message in the logs:

```
WARNING: io.vertx.redis.client.impl.RedisClient - No handler waiting for message: [message, text of the message]
```

### Steps to reproduce

1. Initialize `RedisClient`
2. call `vertx.eventBus().<JsonObject>consumer("io.vertx.redis.channelName").handler(msg -> { /* do something */ });`
3. call `redisClient.subscribe("channelName", res -> { /* verify success */ });`
4. publish some message to `channelName`

You'd expect the handler set in step (2) to be called, but that never happens.

PR #191, opened 2019-09-23
Reinstate pub/sub forwarding to the event bus

guss77:
A possible fix to issue #188 - though much more of a hack than the original approach in vertx-redis-client 3.6.

Even if this PR is not accepted (which I fully expect it will not) I'd appreciate if a developer is willing to discuss how this feature should be re-implemented after its removal in 3.7.


pmlopes:
Hi, there's 2 cases to be addressed in pub/sub

https://github.com/vert-x3/vertx-redis-client/blob/3.6/src/main/java/io/vertx/redis/impl/AbstractRedisClient.java

We can add this back and your or is on the right track.

Issue #183, opened 2019-08-31
Redis connection URI support


Gaket:
### Version

* vert.x core: 3.8.1
* vert.x redis client: 3.8.1

### Context

I've started learning Vert.x, mostly I have Android dev experience.

Trying to connect a Redis instance to Vert.x Verticle. The most convenient way for me would be to simply user Redis URI, as the first variant here, in another Redis library: https://github.com/lettuce-io/lettuce-core/wiki/Redis-URI-and-connection-details

However, I can't find such an option neither in `RedisOptions`, neither in docs: http://vertx.io/docs/vertx-redis-client/java/

Does this library really not have possibility of using a connection URI?

pmlopes:
In master there's support for redis uri, however there's no standard but as master hasn't been released in a final release we could verify and see if we could align the format.

Gaket:
Actually, URI string support was added in Redis 4.0.3. Here you can find [a commit message](https://raw.githubusercontent.com/antirez/redis/4.0/00-RELEASENOTES) (try looking for "-u" in the log), and [here](https://redis.io/topics/rediscli) you can find a description.

I will be happy to contribute, adding a `RedisOptions.setConnectionUri(String connectionUri)` that will parse the URI through `URI.create()` and set according fields of the `RedisOptions`. Also, addition of such a constructor may be considered as well, giving the ease of using a ready URI, often provided by Redis cloud hosts.

Of course, tests will be added as well.

pmlopes:
@Gaket great, please do and ping me if there's any need issue. Note that RedisURI is in the `impl` package so feel free to modify it to properly implement the URI as redis 4.0.3 specifies.

Gaket:
Sorry, had a couple of busy weeks. Going to implement it this Sunday / next Monday. Just for your information.

Gaket:
In RedisURI I renamed the `address` field to `connectionString`, similar to the way it is called in MongoDb: https://docs.mongodb.com/manual/reference/connection-string/

There is a `RedisOptions.setEndpoint(String endpoint)` method that actually behaves (after a few small touches) as a `RedisOptions.setConnectionStringUri(String connectionStringUri)`.

1) Can I rename this public api parameter name and it's javadoc? 

2) Should I rename the method itself or should I create a new method with the proposed name (`RedisOptions.setConnectionStringUri(String connectionStringUri)`) but actually doing the same thing? Or let's just update the Javadoc?

pmlopes:
[...]

PR #193, opened 2019-09-25
Add Redis connection string URI support

Gaket:
Added Redis connection string URI support according to discussion in the issue:
https://github.com/vert-x3/vertx-redis-client/issues/183

Overall, the changes are minimalistic: fixed support of password extraction from the uri (as it's a common thing not to have a user name in this kind of strings).

Please. review the PR commit by commit - the main changes are in the first one, while the others are rather renamings / refactoring.

The overall schema can be found at "Host, port, password and database" block at: 
https://redis.io/topics/rediscli 

There are some open questions:

There is a `RedisOptions.setEndpoint(String endpoint`) method that actually behaves (after a few small touches) as a `RedisOptions.setConnectionStringUri(String connectionStringUri)`.

Should I rename the method itself or should I create a new method with the proposed name (`RedisOptions.setConnectionStringUri(String connectionStringUri)`) but actually doing the same thing? Or let's keep it as it?

### DIFF
--- a/src/main/generated/io/vertx/redis/client/RedisOptionsConverter.java
+++ b/src/main/generated/io/vertx/redis/client/RedisOptionsConverter.java
@@ -1,9 +1,7 @@
 package io.vertx.redis.client;
 
-import io.vertx.core.json.JsonObject;
 import io.vertx.core.json.JsonArray;
-import java.time.Instant;
-import java.time.format.DateTimeFormatter;
+import io.vertx.core.json.JsonObject;
 
 /**
  * Converter and mapper for {@link io.vertx.redis.client.RedisOptions}.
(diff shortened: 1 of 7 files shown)


pmlopes:
Hi, there are a couple of small notes. the IANA registry has 2 entries:

* http://www.iana.org/assignments/uri-schemes/prov/redis
* http://www.iana.org/assignments/uri-schemes/prov/rediss

I believe both should be allowed, so there should be a flag to tell if it's ssl or not.

It also states that query is client specific and there are well known query params that could be present:

* db
* password
* ssl

should we ignore the query part I think it should clearly documented that we do so and that is fine per the official spec.

Gaket:
Thanks for pointing the docs, I think we should better update it. I think I'll have some free time only Sunday evening, October 29. Do we have some time before the next release or should I try to find some time slot earlier?

PR #194, opened 2019-10-10
Add redis URI support according to IANA

Gaket:
Added Redis URI support according to IANA spec for the issue: https://github.com/vert-x3/vertx-redis-client/issues/183

Overall, the changes are minimalistic: fixed support of password extraction from the uri (as it's a common thing not to have a username in this kind of strings).

The overall schema can be found at https://www.iana.org/assignments/uri-schemes/prov/redis as pointed by @pmlopes 

In a separate commit, replaces `addEndpoint` calls with `addConnectionString` which does the same thing but highlights that a connection URI can be passed as a param. This change might be reverted if needed.

### DIFF
--- a/src/main/generated/io/vertx/redis/client/RedisOptionsConverter.java
+++ b/src/main/generated/io/vertx/redis/client/RedisOptionsConverter.java
@@ -1,9 +1,7 @@
 package io.vertx.redis.client;
 
-import io.vertx.core.json.JsonObject;
 import io.vertx.core.json.JsonArray;
-import java.time.Instant;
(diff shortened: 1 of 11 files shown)


Gaket:
I am not sure if the library supports SSL internally and what's required for that. As such, I haven't added `rediss` scheme or `ssl` query parameter support. The latter is also client-specific and not a standard one:
https://iana.org/assignments/uri-schemes/prov/rediss

pmlopes:
It does support ask, it's just an option in the connection

Issue #198, opened 2019-10-31
Movable key commands on cluster client.


wem:
### Version

* vert.x core: 3.8.3
* vert.x redis client: 3.8.3

### Context
Is the reason the clustered client can not handle some kind of commands like EVAL, GEO... (those, they are flagged as movable), that they can contain multiple keys and therefore maybe needs to get validated against the target hash slot?

pmlopes:
Eval for example must be run on all nodes. The "fix" would be to write a custom handle for it like:

https://github.com/vert-x3/vertx-redis-client/blob/master/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java#L70

wem:
Are you sure that EVAL must run against all nodes? The Redis cluster spec says:

> Implemented subset
> Redis Cluster implements all the single key commands available in the non-distributed version of Redis. Commands performing complex multi-key operations like Set type unions or intersections are implemented as well as long as the keys all hash to the same slot.
> 
> Redis Cluster implements a concept called hash tags that can be used in order to force certain keys to be stored in the same hash slot. However during manual resharding, multi-key operations may become unavailable for some time while single key operations are always available.
...
And
...
> Since cluster nodes are not able to proxy requests, clients may be redirected to other nodes using redirection errors -MOVED and -ASK. The client is in theory free to send requests to all the nodes in the cluster, getting redirected if needed, so the client is not required to hold the state of the cluster. However clients that are able to cache the map between keys and nodes can improve the performance in a sensible way.

As i understand (and tested with near all of the movable flagged commands and https://github.com/Grokzen/docker-redis-cluster) commands with mutliple keys they target different hash slots 
will fail anyway with "(error) CROSSSLOT Keys in request don't hash to the same slot" like `redis-cli -h localhost -p 7000 "redis.call('SET', KEYS[1], ARGV[1]) redis.call('SET', KEYS[2], ARGV[1])" 2 {hash_tag}.key {other-hash_tag}.key value`.

So i think those commands can be implemented straight forward, ideal but optional with a validation of the hash slots before command execution. Or i am wrong?

pmlopes:
Hi, `eval` is not on the list:

https://github.com/vert-x3/vertx-redis-client/blob/3.8/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java#L124-L129

[...]

PR #199, opened 2019-11-03
Support for movable keys commands on clustered Redis client (send and batch)

wem:
https://github.com/vert-x3/vertx-redis-client/issues/198

- Extraction of the keys on movable key commands.
- Validation of those keys that they target the same hash slot.
- Normalization of hash slot issue related exception messages.
- Consider commands (currently only WAIT) they must be run against master nodes.

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/KeyExtractor.java
+++ b/src/main/java/io/vertx/redis/client/impl/KeyExtractor.java
@@ -1,0 +1,142 @@
+package io.vertx.redis.client.impl;
+
+import java.nio.charset.StandardCharsets;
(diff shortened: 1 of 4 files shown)


wem:
Unfortunately, there is a missing point i did found. WAIT needs to run on master at any time ... I will adjust the PR to consider this.

pmlopes:
@wem would you please port this PR to master too? I can't merge it as is. If you make the second PR to master then I'll merge them both! :+1:

wem:
@pmlopes Done, thx!
The PR to master contains an additional fix for batches, as the slot check did fail if the first command was a single key command.

Issue #201, opened 2019-11-11
Support cluster commands on redis cluster client


wem:
### Version

* vert.x core: 3.8.3
* vert.x redis client: 3.8.3

Cluster commands (at least the query variants e.g. slots, nodes) can become very important to be available in this mode (e.g. to implement a distributed lock). Unfortunately, the cluster command is currently marked as unsupported in general. IMHO are those command variants more "advanced" and users they use them, know what they do and i think the query variants have no risk from the sight of the cluster. So, would it be possible to remove the cluster command from the list of unsupported commands? If that's the case, i will do a PR.

pendula95:
Yes, this was a mistake made by me. CLUSTER command should be supported and is only a matter of removing it from LIST of unsupported.

pmlopes:
the cluster commands are merged on master

PR #202, opened 2019-11-12
Support cluster command on RedisClusterClient.

wem:
https://github.com/vert-x3/vertx-redis-client/issues/201


wem:
This looks like a brittle test, as the change has no relation to the failed test.

PR #203, opened 2019-11-16
Fixes #187

### LINKED ISSUE #187: Context difference
### Version

* vert.x core:4.0.0-milestone2
* vert.x redis 4.0.0-milestone2

### Context

When acquiring a connection from Redis client the handler is called on a different context instead of the context used when calling.

Is that intentional? 

Do we need to execute runOnContext inside the handler in order to return back to the calling context?

zenios:
Signed-off-by: zenios <dimitris.zenios@gmail.com>

Since we don't support shared client we assume that all methods of RedisConnectionImpl will be called either from the context that initialised the RedisClient or the context of io.vertx.core.http.impl.pool

In either case both contexts share the same event loop so no synchronisation issues. This is what is stated on the documention of pool

> A connection is delivered to a {@link Waiter} on the pool's context event loop thread, the waiter must take care of calling {@link io.vertx.core.impl.ContextInternal#emitFromIO} if necessary.

Having that in mind content.runOnContext (Context here is the context used when creating the client)  should only be used in cases where a call to RedisConnectionImpl  arrived from the pools context.

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisConnectionImpl.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisConnectionImpl.java
@@ -4,8 +4,6 @@ import io.vertx.core.*;
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.eventbus.EventBus;
 import io.vertx.core.http.impl.pool.ConnectionListener;
-import io.vertx.core.impl.ContextInternal;
-import io.vertx.core.impl.PromiseInternal;
 import io.vertx.core.impl.logging.Logger;
 import io.vertx.core.impl.logging.LoggerFactory;
 import io.vertx.core.json.JsonObject;
@@ -27,7 +25,6 @@ public class RedisConnectionImpl implements RedisConnection, ParserHandler {
   private static final ErrorType CONNECTION_CLOSED = ErrorType.create("CONNECTION_CLOSED");
 
   private final ConnectionListener<RedisConnection> listener;
-  private final ContextInternal context;
   private final EventBus eventBus;
   private final NetSocket netSocket;
   // waiting: commands that have been sent but not answered
@@ -40,10 +37,9 @@ public class RedisConnectionImpl implements RedisConnection, ParserHandler {
   private Handler<Void> onEnd;
   private Handler<Response> onMessage;
 
-  public RedisConnectionImpl(EventBus eventBus, ContextInternal context, ConnectionListener<RedisConnection> connectionListener, NetSocket netSocket, RedisOptions options) {
+  public RedisConnectionImpl(EventBus eventBus, ConnectionListener<RedisConnection> connectionListener, NetSocket netSocket, RedisOptions options) {
     this.listener = connectionListener;
     this.eventBus = eventBus;
-    this.context = context;
     this.netSocket = netSocket;
     this.waiting = new ArrayQueue(options.getMaxWaitingHandlers());
     this.recycleTimeout = options.getPoolRecycleTimeout();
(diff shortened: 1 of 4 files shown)


src/main/java/io/vertx/redis/client/impl/RedisConnectionImpl.java
```diff
@@ -110,19 +111,16 @@ public RedisConnection send(final Request request, Handler<AsyncResult<Response>
 
     // encode the message to a buffer
     final Buffer message = ((RequestImpl) request).encode();
-    final ContextAwareHandler<AsyncResult<Response>> ctxAwareHandler = new ContextAwareHandler<>(handler);
     // all update operations happen inside the context
-    context.runOnContext(v -> {
```
pmlopes: Tests are failing and i believe this is the root cause. The queue implementation is not synchonized to make it fast so all mutations must happen on the event loop. Also doing that + socket writes ensures the order is preserved and correct for the callbacks.

A short optimization here is to verify that were on the same context as the socket and avoid the run on context.

src/main/java/io/vertx/redis/client/impl/RedisConnectionImpl.java
```diff
@@ -236,75 +229,57 @@ public void handle(Response reply) {
       return;
     }
 
-    // all update operations happen inside the context
-    context.runOnContext(v -> {
```
pmlopes: Same as previous comment.

src/main/java/io/vertx/redis/client/impl/RedisConnectionImpl.java
```diff
@@ -314,25 +289,12 @@ public void fatal(Throwable t) {
...
-    // call the exception handler if any
-    if (onException != null) {
-      onException.handle(t);
-    }
+    fail(t);
   }
 
   private void cleanupQueue(Throwable t) {
-    // all update operations happen inside the context
-    context.runOnContext(v -> {
```
pmlopes: Same as previous comment.

src/main/java/io/vertx/redis/client/impl/RedisConnectionImpl.java
zenios: Why do we need synchronization?

As explained above all methods should be called from the same event loop.

Please correct me if i am wrong

zenios: Could you please tell me which tests are failing for you?

pmlopes: https://travis-ci.org/vert-x3/vertx-redis-client/builds/612915093

zenios: onConnect callback that creates the netClient was called without a context.

That resulted to each netClient instance being executed on a different context and hence different thread.

In the case of clustered tests, 6 connection where initialised on 6 different threads

In order to fix the issue i either had to:
- synchronise the queue (Either by synchronised or runOnContext)
- Run the netClient on the same context as the the RedisClient.create(vertx)

I chose the later, but with some very simple benchmarks on the intensive test cases (RedisClusterTest.runTheSlotScope) this implementation takes on average 200ms more to complete

I also checked vertx-sql-client and there the netClient is also created on the same context as the PgClient.create(vertx)

zenios: The tests failing on Sentinel is coincidental. The first commit that fixed the problem passed the tests just fine but the second commit witch only added some tests failed on sentinelTest (A test that was not changed between the first and the second commit)

PR #212, opened 2020-02-19
Backported the modern client to the 3.9 branch and fully deprecated t…

pmlopes:
…he existing client

Motivation:

The current 3.x client is to be replaced on 4.0. It is dated and suffers from lots of issues. This PR will expose the new client in the 3.x branch and will live side by side with the old API with enough warnings to move to the new API.

The new client supports any connection style, single, sentinel and cluster, plus supports all existing redis command and new ones as well as extensions.


underkin:
Paolo, 
Am I correct understanding that  this branch is using the new connection style used in vertx 4.0 redis library. But it still keeps the  RedisClient, so code for vertx3 will work as before even if the underlying implementation is different?

Also, will  the new redis library in vertx4 change the API or will we still have the same  RedisClient api we used in vertx3?

pmlopes:
This branch has the new client backported to the 3.x branch. The old api is still present as is a shim on top of the new client (however it has edge cases and limitations, that's why it's deprecated)

The goal is to expose the new api which is very different as soon as possible to the users.

I don't expect much more changes upstream so the api is quite stable atm.

underkin:
Ok. I thought there would be kind of a compatibility wrapper, so the API does not change for current vertx3 users, even if the underlying code change so much.

pmlopes:
The underlying code changed, because it's now just delegating to the new client. The reason for this is that as the old API is to be removed on 4.0 we allow users to start migrating early and help finding any new bugs that could arise from this.

underkin:
Sounds great. Will start trying out once merged

PR #236, opened 2020-05-22
encode the password and test RedisURI for encode password

### LINKED ISSUE #235: wrong password with old api in 3.9
In the old api, it will convert `io.vertx.redis.RedisOptions` to a connectionString like:

```
redis://localhost:6379/0?password=xxx
```

but if my password contains '#', it will get a wrong password in `RedisURI`, it seem like that the password dos not encode. e.g

```
redis://localhost:6379/0?password=abc#123
```
The real password is `abc#123`，but the param in the query is `abc`

okou19900722:
fixed #235

### DIFF
--- a/src/main/java/io/vertx/redis/impl/RedisClientImpl.java
+++ b/src/main/java/io/vertx/redis/impl/RedisClientImpl.java
@@ -29,9 +29,10 @@ import io.vertx.redis.RedisOptions;
 import io.vertx.redis.RedisTransaction;
 import io.vertx.redis.Script;
 import io.vertx.redis.client.*;
-import io.vertx.redis.client.impl.types.MultiType;
 import io.vertx.redis.op.*;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLEncoder;
 import java.util.*;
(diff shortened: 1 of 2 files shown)


src/test/java/io/vertx/redis/client/impl/RedisURITest.java
```diff
@@ -33,6 +36,14 @@ public void testTwoPasswordsAreGiven() {
     Assert.assertEquals("Password is not correct", "pass", redisURI.password());
   }
 
+  @Test
+  public void testPasswordWithEncodeInQuery() throws UnsupportedEncodingException {
```
gaol, line 40: This test passes even without the fix. Actually I believe this is not a bug.

okou19900722: have you run the test? a new test was commit.you can review it

gaol: After contact offline, I understand the problem now. Yes it should be a bug in the old API :)

pmlopes:
Thanks!

PR #247, opened 2020-09-08
Issues/246

### LINKED ISSUE #246: Array,Set issue with RESP3 when length is even
Since list and sets are parsed as multitype, based on the following line

https://github.com/vert-x3/vertx-redis-client/blob/e9e3c0db349d662e676c04235841c68488d6f6d2/src/main/java/io/vertx/redis/client/impl/types/MultiType.java#L35

each of the above that happens to have even number of elements, is considered a map resulting to each odd (indexed) element to be lost

pmlopes:
Motivation:

Fixes #246 

Will handle Map only as Map, Multi will have "slow" fallback for map related operations as a convenience that was available on 3.x and is also implied on the redis docs.

### DIFF
--- a/src/main/java/io/vertx/redis/client/Response.java
+++ b/src/main/java/io/vertx/redis/client/Response.java
@@ -222,6 +222,16 @@ public interface Response extends Iterable<Response> {
   }
 
   /**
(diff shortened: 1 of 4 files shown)


pmlopes:
@zenios can you give it a look?

PR #251, opened 2020-10-01
Cluster password

pmlopes:
Motivation:

Ensure that we work as expected with password protected clusters.


pmlopes:
@pendula95 your setup seems to work :+1: Thanks!

Issue #264, opened 2020-11-23
Possibly Not Setting ConnectionPool Max Queue Size correctly 


calebvarberg:
### Version

3.9.2 (appears to still be an issue)

### Context

I recently noticed a lot of these exception from code I was running 

> `timestamp=12:00:24.357, level=WARN , loggingId=, thread=vert.x-eventloop-thread-0, logger=s.service, message=key=key, correlationId=17d90dc2-1138-4818-ab5c-7faf5f3d79e7, exceptionClass=ConnectionPoolTooBusyException, exceptionMessage=Connection pool reached max wait queue size of 6,
> io.vertx.core.http.ConnectionPoolTooBusyException: Connection pool reached max wait queue size of 6
	at io.vertx.core.http.impl.pool.Pool.lambda$nextTask$1(Pool.java:293)
	at io.vertx.core.http.impl.pool.Pool.checkPendingTasks(Pool.java:250)
	at io.netty.util.concurrent.AbstractEventExecutor.safeExecute(AbstractEventExecutor.java:164)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:472)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:497)
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989)
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
	at java.lang.Thread.run(Thread.java:748)`

I thought this was strange as I had set the `maxPoolWaiting` in RedisOption to be a number larger than 6. Upon further inspection I found that the code to create the Connection Pool in `RedisConnectionManager.java` sets the MaxWaitQueue size to be equals to `RedisOptions.getMaxPoolSize()`, and it sets the `maxWeight` of the pool to be `RedisOptions.getMaxPoolWaiting()`.

This feels backwards to me, and those fields are being used the opposite of how i would have thought from reading the [docs](https://vertx.io/docs/apidocs/io/vertx/redis/client/RedisOptions.html). Which seem to suggest `maxPoolWaiting` should be the size of the queue, and `maxPoolSize` would be the maximum amount of connections.

calebvarberg:
[...]

PR #265, opened 2020-11-24
BUG-264 updating the connection pool creation for the redis client

calebvarberg:
Motivation:

attempting to fix the "bug" described in https://github.com/vert-x3/vertx-redis-client/issues/264


pmlopes:
Thanks, indeed the values are used in the wrong order! :+1:

PR #271, opened 2020-12-10
Backport redis cluster authentication changes

### LINKED ISSUE #181: Redis Cluster always connected to 127.0.0.1
Hi Vertx Devs,

when I try following code with vertx 3.8.0:

Vertx vertx = Vertx.vertx(); RedisClient redis = RedisClient.create(vertx, new RedisOptions().setSentinels(List.of("192.168.5.20:8400", "192.168.5.20:8401", "192.168.5.20:8402")).setMasterName("MASTER_NAME")); Future<JsonArray> future = Future.future(); redis.keys("*", future); future.setHandler(res -> { if (res.succeeded()) { System.out.println(res.result()); } else { res.cause().printStackTrace(); } });
it always throw an following exception:

`
io.netty.channel.AbstractChannel$AnnotatedConnectException: Connection refused: /127.0.0.1:6379
at java.base/sun.nio.ch.SocketChannelImpl.checkConnect(Native Method)

`

But actually nothing is listening on 127.0.0.1:6379
Maybe its an internal issue ...

 @gillbates

### LINKED ISSUE #270: Cluster client is only authenticating initial connection
### Version

3.9.4

### Context

We are upgrading to vertx 3.9.4 and ported to the new redis client implementation.

Connecting to an insecure cluster works, but requests to an authenticated cluster fail with:

`NOAUTH Authentication required.`

When inspecting the netty logs, it's apparent that only the first connection sends a `AUTH` command using the configured password. That connection is used to get cluster/slots information and is dropped afterwards.
The cluster client then establishes connections to each discovered cluster member, but does **not** send an `AUTH` command after connecting.

I think the reason is, that after determining all available slots, the cluster client connects using the endpoints from the `SLOTS` response, which do not include the password.

A fix could be to patch derived endpoints, if the initial connection string included a password.

### Do you have a reproducer?

Enable log activity on the net client and connect to a redis cluster that required authentication, for example ElastiCache.

### Steps to reproduce

see above

### Extra

* n/a

pyranja:
Motivation:

When connecting to a redis cluster, the client now propagates username
and password from the initial connection to each of the cluster
member connections.

This is a back port of changes to `RedisURI` and `Slots` from the master branch.

Closes #270  

Conformance:

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
@@ -148,7 +148,7 @@ public class RedisClusterClient implements Redis {
       final RedisConnection conn = getConnection.result();
 
       // fetch slots from the cluster immediately to ensure slots are correct
-      getSlots(conn, getSlots -> {
+      getSlots(endpoints.get(index), conn, getSlots -> {
         if (getSlots.failed()) {
           // the slots command failed.
           conn.close();
@@ -217,7 +217,7 @@ public class RedisClusterClient implements Redis {
     connectionManager.close();
   }
 
-  private void getSlots(RedisConnection conn, Handler<AsyncResult<Slots>> onGetSlots) {
+  private void getSlots(String endpoint, RedisConnection conn, Handler<AsyncResult<Slots>> onGetSlots) {
 
     conn.send(cmd(CLUSTER).arg("SLOTS"), send -> {
       if (send.failed()) {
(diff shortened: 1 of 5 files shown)


pyranja:
I'm afraid the travis build will not run anymore.

When building locally `acquireConnectionsTest(io.vertx.redis.client.test.RedisPooledTest)` fails.
That test also fails on the 3.9 branch for me, and also failed in the last travis build - I don't think my changes are affecting it.

@pmlopes what are the chances to get a 3.9.5 release including these changes?
I assume that most effort will be spent on the 4.x releases?

PR #276, opened 2021-01-29
Check total endpoints against the right value

### LINKED ISSUE #274: Cluster Pool Size Error
Hi
The validation agains the maxPoolSize as described in the link, should be done agains the number of endpoints rather than the number of slots. The reason is that the number of slots can be larger than the number of nodes.
https://github.com/vert-x3/vertx-redis-client/blob/c2eeb3d3c322b3369d3a600149c1591e762fdce1/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java#L164

pmlopes:

Motivation:

Fix #274 : The validation was asserting the wrong value


orxelm:
Thanks 👍

Issue #279, opened 2021-02-15
Add automated tests for TLS


pmlopes:
Currently we have no automated tests for TLS, we should use the docker image:

https://hub.docker.com/r/runnable/redis-stunnel/

for quickly assert that TLS works.

* Direct connection with `rediss`
* Upgrade connection if `ssl` is configured manually in the net client options

pmlopes:
This issue should then be backported to 3.9

PR #280, opened 2021-02-15
Ensure we work in TLS

pmlopes:

Motivation:

Ensure the client works with TLS connections


pmlopes:
Fixes #279

PR #282, opened 2021-03-03
Adapt the client to the internal pool changes

vietj:


pmlopes:
@vietj I think there's a behavior change in the new pool, the test:

```java
  @Test(timeout = 30_000L)
  public void acquireConnectionsTest(TestContext should) {
    final Vertx vertx = rule.vertx();
    final Async test = should.async();

    Redis client = Redis.createClient(
      vertx,
      new RedisOptions()
        .addConnectionString("redis://localhost:7006")
        .setMaxPoolSize(10)
        .setMaxPoolWaiting(10));

    final AtomicInteger counter = new AtomicInteger();

    // this test asserts that the pools behaves as expected it shall return 10 new connections
    // and will fail on the 21st call as the 10 waiting slots are taken

    vertx.setPeriodic(500, event -> {
      counter.incrementAndGet();
      client.connect(event1 -> {
        if (event1.succeeded()) {
          System.out.println(counter.get());
          should.assertTrue(counter.get() <= 10);
        } else {
          System.out.println(counter.get());
          should.assertTrue(counter.get() == 21);
          vertx.cancelTimer(event);
          vertx.runOnContext(v -> test.complete());
        }
      });
    });
  }
```

Is now failing, while it wasn't before. The test was to check that if the configuration was working as expected, this is what it does:

```java
pool = new Pool<>(
        ctx,
        connectionProvider,
        10,
        1,
        10,
        c -> {},
        c -> {},
        false);
```

Before it would get 10 connections right away, the next 10 would queue and after that it would fail. Now it always succeeds. I've stopped the test after 50 so I assume the weight calculation is different somehow.

PR #286, opened 2021-03-22
Fix wrong connection validation

dangfan:
Currently, when a connection [is returned to the pool](https://github.com/eclipse-vertx/vert.x/blob/358cee178d9e1754ae57b5f8e95b605a74be22d6/src/main/java/io/vertx/core/net/impl/clientconnection/Pool.java#L304), the validation is always `false` due to the `expirationTimestamp` in `RedisStandaloneConnection` is always 0.

The `recycle` function indicates that if the `expirationTimestamp` should be 0 if we don't want to recycle, therefore, I changed the `isValid()`. However, since `recycle()` is never called, and this client uses a timer to close connections, I wonder it might be better to remove `expirationTimestamp`, `recycle()`, and always return `true` in `isValid()`.

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisStandaloneConnection.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisStandaloneConnection.java
@@ -20,7 +20,6 @@ import io.vertx.redis.client.impl.types.Multi;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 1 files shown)


pmlopes:
@dangfan I don't think returning always `true` is correct. Maybe the correct behavior should be:

in the constructor of Standalone Connection:

```java
expirationTimestamp = recycleTimeout > 0 ? System.currentTimeMillis() + recycleTimeout : 0L;
```

Which means when the connection is returned to the pool if it's older than the expiration timeout now it will return false from `isValid()`

Also `isValid()` should return `false` if any of the `fail()` or `fatal()` have been called because this means the internal REDIS parser is out of sync and cannot recover.

WDYT?

dangfan:
@pmlopes I agree with you that `isValid()` should return `false` if any of the `fail()` or `fatal()` has been called.

As for the `expirationTimestamp`, I assume that this is designed for closing idle connections. If my assumption is correct, the timer in `RedisConnectionManager` (L47) should just work, and this `expirationTimestamp` would be useless.

pmlopes:
@dangfan would you like to address the issues discussed here in the PR? Let me try to summarize them:

1. `isValid()` returns `false` if either `fail()` or `fatal()` have been called, otherwise `true`.
2. `recycle()` can be safely removed as it's dead code
3. remove any other unused bits, for example `expirationTimestamp` may not be used anymore at the `RedisStandaloneConnection`
4. In order for the idle reaper to be executed the `RedisConnectionManager` expects that the timer created in the `start()` method is running. We need to ensure that that method has been called, or perhaps make it part of the constructor, so we're always certain.

Did I miss anything?

dangfan:
@pmlopes your summary looks good to me. I'll try to fix them in the following couple of days.

dangfan:
@pmlopes I've finished and please review the commits.
P.S., the CI failed due to `Error: Resource not accessible by integration`. I assume that this is not what I can fix.

pmlopes:
@dangfan thanks!

kalaibabu:
Hi ,

I have referred to the vertx-rx-java-3.8.0.jar method below
imports  io.vertx.rxjava.redis
setBinaryWithOptions - vert.x.3

Vert.x.4 , vertx-rx-java-4.1.2.jar contains what will be considered equivalent methods

Issue #255, opened 2020-10-13
"Implementing Reconnect on Error" example in README causes a connection leak


koby08:
Hi,

It seems that the "Implementing Reconnect on Error" section in the [official docs](https://vertx.io/docs/vertx-redis-client/java/) for this client contains a bad example. If someone runs this example, each time it reconnects, the old connection will remain open. This will cause a connection leak to redis server.

The reason for this leak is that in each reconnect attempt Redis.createClient() is called, but without closing the old one. Please notice that calling client.close() won't help as well, because it is an instance of a connection from the connection pool.

In order to fix this, you need to call close() on the instance returned from Redis.createClient() - which will close the connection pool.

kobytg17:
Actually another issue with the current implementation is if there are several in-flight requests to Redis, and all fire an exception due to some failure, there will be multiple connection attempts that will create dangling clients

Stef16Robbe:
Think this can be closed since https://github.com/vert-x3/vertx-redis-client/pull/292 and https://github.com/vert-x3/vertx-redis-client/pull/377 have been merged

PR #292, opened 2021-05-21
Fix cluster reconnect example

pendula95:
Motivation: Current example shows bad practice that can lead to dangling connections and concurrent issues.
It should tackle the issue mentioned #255


pmlopes:
I've cherry picked the commit to `master` as there's too much noise in the PR. The commit retains full attribution.

https://github.com/vert-x3/vertx-redis-client/commit/790475d585fe13528c13f87c6629eaadf7bfbe66

pendula95:
I think you are right. Looking at my own implementations in projects I see `((RedisClusterConnection) redisConnection.getDelegate()).close();` which will call https://github.com/vert-x3/vertx-redis-client/blob/3290a06ac1ba0b3a2b3f9897aafcc607ca37c50b/src/main/java/io/vertx/redis/client/impl/RedisClusterConnection.java#L522

The example should be changed to feature this.

PR #319, opened 2022-01-20
Adding and example of client handling credentials - user and password 

gutmox:
Adding and example of client handling credentials - user and password 

Motivation:

When you encode passwords and pass then as part of the connection string, redis clusters don't normally decode them back and returns a bad credentials. This example shows how to pass user as query param and password as part of Options.  

Conformance:

### DIFF
--- a/src/main/java/examples/RedisExamples.java
+++ b/src/main/java/examples/RedisExamples.java
@@ -212,4 +212,22 @@ public class RedisExamples {
         conn.send(Request.cmd(Command.GET).arg("key"));
       });
   }
+
+  public void example14(Vertx vertx) {
+    ;
+    Redis.createClient(
+        vertx,
+        // The client handles using credentials - user and password - with
+        // complex non-alphanumerical passwords, when .
+        // when you encode passwords and pass then as part of the connection string,
+        // redis clusters don't decode them back and returns a bad credentials
+        // error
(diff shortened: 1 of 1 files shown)


src/main/java/examples/RedisExamples.java
```diff
@@ -212,4 +212,22 @@ public void example13(Vertx vertx) {
...
   }
+
+  public void example14(Vertx vertx) {
+    ;
+    Redis.createClient(
+        vertx,
+        // The client handles using credentials - user and password - with
+        // complex non-alphanumerical passwords, when .
+        // when you encode passwords and pass then as part of the connection string,
+        // redis clusters don't decode them back and returns a bad credentials
```
pmlopes, line 223: Isn't this a bug instead of a feature we want to document?

gutmox: Yes, 🤔, that's right - I just wanted to share something I took me some time to figure out with the current version. 

I'll prepare a PR with a proposal to parse the connection string param instead using URI

Anyway, I think it's still a valid case, combining passing user as query param and password via Options field. Maybe changing my comment to make it more neutral ?.

gutmox:
Checked that uri.getUserInfo() actually decode username & password. Passing encoded values should work well.

PR #328, opened 2022-05-10
use the new key resolution algorithm

pmlopes:

Motivation:

Redis 7 introduced a new way to perform key resolution. This PR removes the wrong way we were using and follows the requirements by redis itself.


liuchong:
Hi, what is the `needGetKeys` for? @pmlopes Thanks!

liuchong:
Another question, look at the below commands, are they `readOnly` or not?
```
Command CONFIG = new CommandImpl("config", -2, null, false, false);
Command SAVE = new CommandImpl("save", 1, null, false, false);
```
They are having a `null` readOnly argument by with no keyLocators, and they should by default got a readOnly `true`.

pmlopes:
That information is extracted from the `COMMANDS` command. If a command has a flag `read` or `write` then it's clear if the command is read only or not, but some cases like these don't include any flag, so nothing can be said. Redis states that for these we should rely on the server to decide.

It's not a bug, it's what the metadata tells us. Before we were assuming all commands were either read only or not but it seems our assumption was wrong.

Do not you don't need this metadata unless you want to use a cluster connection. For most cases all you need is the regular command name.

PR #337, opened 2022-05-24
new encode function

Xpecya:
#336 I opened a new PR @pmlopes Thank you

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RequestImpl.java
+++ b/src/main/java/io/vertx/redis/client/impl/RequestImpl.java
@@ -149,6 +149,40 @@ public final class RequestImpl implements Request {
     return encode(Buffer.buffer());
   }
 
+  Buffer encode(String[] args) {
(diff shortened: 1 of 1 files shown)


pmlopes:
@Xpecya give a look at https://github.com/vert-x3/vertx-redis-client/pull/338

This PR may work on the full redis stack API. I'm thinking that using varargs can make it easier for the end user. If we can optimize the encoding to be less GC even better

Xpecya:
@pmlopes  I agree that String[] args is a convinent

Xpecya:
@pmlopes users would more likely to use get/set/incr... instead of send(command, args) or send(request) in my opinion, so it doesn't really matter how the args are. Just make sure the user interfaces(get/set/incr etc) stay the same. I agree that this PR may work on #338, perhaps the first thing is to do some modification on the Request class, store the Object[] directly instead of create a ArrayList first... I'll try later on that branch and open a new PR

PR #346, opened 2022-09-06
Compatible with older versions of auth requests that do not require the user parameter

okou19900722:
Signed-off-by: okou19900722 <ilayan19900722@gmail.com>

Motivation:

Explain here the context, and why you're making that change, what is the problem you're trying to solve.

Conformance:


pmlopes:
Although this works, I believe we should somehow refactor to expose the connection metadata which includes the server version, if I'm not mistaken...

pmlopes:
Oh wait, this is a 3.9 issue, nevermind the previous comment.

PR #352, opened 2022-10-15
Issues/cluster from dns

pmlopes:
Motivation:

Cluster deployments that use the DNS to announce the initial set of servers composing the cluster can now be discovered by the cluster client instead of requiring users to specify the initial list of servers.

This is the case of AWS `memoryDB` for example.

### DIFF
--- a/src/main/java/io/vertx/redis/client/Redis.java
+++ b/src/main/java/io/vertx/redis/client/Redis.java
@@ -42,7 +42,7 @@ public interface Redis {
   }
 
   /**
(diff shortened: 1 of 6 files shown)


OmerShemesh:
Could this be helpful when using ElastiCache cluster with configuration endpoint?

pmlopes:
@OmerShemesh yes, AWS exposes the full cluster nodes in the DNS record of the service, recently they also have support for custom authentication using IAM. This PR is a start prototype to collect the cluster endpoints from the DNS but isn't finished. It should also be guarded by a flag to avoid extra startup delays.

PR #354, opened 2022-10-18
close tainted connection after evicting from pool

### LINKED ISSUE #353: PooledRedisConnection.close() does not close the connection if it is tainted (Pub/Sub)
On closing tainted connections, they are evicted from the connection pool without being closed which causes connection leaks.

**PooledRedisConnection#close()** -> **connection.reset()**:
https://github.com/vert-x3/vertx-redis-client/blob/741f219b1164e36558ff9f0abaa4044b0d1217c9/src/main/java/io/vertx/redis/client/impl/PooledRedisConnection.java#L82-L91

**RedisStandaloneConnection.reset()** -> **evict()**:
https://github.com/vert-x3/vertx-redis-client/blob/741f219b1164e36558ff9f0abaa4044b0d1217c9/src/main/java/io/vertx/redis/client/impl/RedisStandaloneConnection.java#L456-L468

**RedisStandaloneConnection.evict()**:
https://github.com/vert-x3/vertx-redis-client/blob/741f219b1164e36558ff9f0abaa4044b0d1217c9/src/main/java/io/vertx/redis/client/impl/RedisStandaloneConnection.java#L470-L479

`netSocket.close()` is missing.

Affected version: 4.3.4

Eng-Fouad:
Fixes #353

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisStandaloneConnection.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisStandaloneConnection.java
@@ -462,6 +462,7 @@ public class RedisStandaloneConnection implements RedisConnectionInternal, Parse
     // cannot reset if connection is tainted (custom DB/AUTH/PUBSUB)
     if (tainted) {
       evict();
+      forceClose();
       return false;
     }
     return true;


pmlopes (approved):

pmlopes:
Yes, we can be more strict and remove the bad connections from the pool immediately instead of keep them stalling the pool and degrading it, until a timeout occurs.

Issue #365, opened 2023-02-17
Create too many connections to server


haohuif:
### Questions

Reids runs in cluster mode. When userReplicas=always is specified, a large number of connections will be created to the server

After analyzing the source code, I think it is because RequestImpl does not implement the hashCode and equals methods. In the getConnection method of RedisConnectionManager, the request object will be specified to form the ConnectionKey together with the  url string of the endpoint. Therefore, each time the connection is obtained through RedisClusterClient, a new connection will be built because the key does not exist

### Version

4.2.7

PyAntony:
I am facing the same issue; I believe this is a bug. The issue is simply resolved by using`userReplicas=never`, so this is definitely happening when using connections to replica nodes. What I have observed is the following:

1) Connection to replica node is used. This connection uses `readonly` mode (https://github.com/vert-x3/vertx-redis-client/issues/160); if it didn't use `readonly` we would simply get the `MOVED` exception.
2) Client gets response successfully, but after that new connections to all cluster nodes are set. This happens with each and every request (I can see this in the logging by using `quarkus.log.category."io.vertx.redis".level=DEBUG`).
3) On the server side the connections in `readonly` mode are never closed; they keep piling up. We can track this by using the `CLIENT LIST` command. There is a point in which the server won't accept more connections; in my case my application just halted (as discussed here: https://github.com/smallrye/smallrye-reactive-messaging/issues/2102).

So the problems That I see are:
- Why are all connections to the cluster reset after each request to a slave node? Resetting all connections again and again on each request is very expensive. I don't think this is intended isn't?
-  There is definitely an issue with the way the connections are closed (are they even closed???). The server doesn't close these open connections in `readonly` mode, so there is something buggy with the client, maybe a swollen exception somewhere that triggers a recreation of the entire connection pool again and again. @haohuif mentioned that "a new connection will be built", but these new connections built are not coming from the same pool since we would quickly run out of available connections and get the pool exception ("not available connections" or something like that). So my hypothesis is that a new pool is created over and over.

[...]

PR #374, opened 2023-04-07
Adding equals/hashcode to members that compose key for connection pool

pmlopes:
Motivation:

Connection pool builds a key based on connection url and setup command, this PR adds equals and hashcode implementation to objects that are used to compose that key to avoid uneeded generation of connections.


pmlopes:
Fixes #365

PR #392, opened 2023-06-20
Implement tracing and client metrics

### LINKED ISSUE #357: Implement io.vertx.core.spi.metrics.ClientMetrics

Ladicek:
Motivation:

- make sure that the Redis client calls back on the original context
  - this used to be the case for the clustered clients, but nor for the standalone client, which is fixed here
- make connection pooling explicit in the code
  - the internal structure of connection handling is more obvious by using the `PooledRedisConnection` type explicitly
  - this even revealed a bug in `RedisSentinelClient` which would lead to class cast errors in some failure scenarios
- implement tracing and client metrics
  - this is done similarly to how the Vert.x SQL client does it, which is a little less precise, but easier to understand

Resolves #357

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisConnectionInternal.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisConnectionInternal.java
@@ -15,6 +15,9 @@
  */
 package io.vertx.redis.client.impl;
 
+import io.vertx.core.impl.VertxInternal;
+import io.vertx.core.spi.metrics.ClientMetrics;
+import io.vertx.core.tracing.TracingPolicy;
 import io.vertx.redis.client.RedisConnection;
 
 public interface RedisConnectionInternal extends RedisConnection {
@@ -27,4 +30,28 @@ public interface RedisConnectionInternal extends RedisConnection {
    * Returns {@code true} is this connection can be reset. This means that the connection didn't enter PubSub mode.
    */
   boolean reset();
+
+  /**
+   * Returns whether this connection is "tainted". A connection is called tainted if it changes the default state,
+   * for example, when a connection enters pub sub mode, or specific features are activated such as changing a database
+   * or different authentication is used.
+   */
+  boolean isTainted();
+
+  VertxInternal vertx();
+
+  /**
+   * Returns the {@linkplain RedisURI URI} of the Redis server to which this connection is connected.
+   */
+  RedisURI uri();
+
+  /**
+   * Returns the {@linkplain ClientMetrics client metrics} for this connection.
+   */
+  ClientMetrics metrics();
+
+  /**
+   * Returns the {@linkplain TracingPolicy tracing policy} configured for this connection.
+   */
+  TracingPolicy tracingPolicy();
 }
--- a/src/test/java/io/vertx/redis/client/test/RedisClusterTest.java
+++ b/src/test/java/io/vertx/redis/client/test/RedisClusterTest.java
@@ -1213,4 +1213,13 @@ public class RedisClusterTest {
         });
       });
   }
+
+  @Test
+  public void preservesContext(TestContext should) {
+    PreservesContext.sendWithoutConnect(client, should);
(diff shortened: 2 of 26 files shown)


Ladicek:
If you agree, I'll submit a backport PR against the `4.x` branch after this is merged.

Ladicek:
No idea why the action failed -- the _Run tests_ step succeeds, but then _Publish Unit Test Result_ fails with `Error: Resource not accessible by integration` :shrug:

src/main/java/io/vertx/redis/client/impl/RedisConnectionInternal.java
```diff
@@ -27,4 +30,14 @@ public interface RedisConnectionInternal extends RedisConnection {
    * Returns {@code true} is this connection can be reset. This means that the connection didn't enter PubSub mode.
    */
   boolean reset();
+
+  boolean isTainted();
```
vietj, line 34: javadoc :-)

src/test/java/io/vertx/redis/client/test/RedisClusterTest.java
```diff
@@ -1213,4 +1213,13 @@ public void testCommandWithoutReadOrWrite(TestContext should) {
         });
       });
   }
+
+  @Test
+  public void preservesContext(TestContext should) {
```
vietj, line 1218: good idea

src/main/java/io/vertx/redis/client/impl/RedisConnectionInternal.java
vietj: thanks, I know it was there before, but I think it is important to define what it means clearly

vietj:
@Ladicek can you rebase your PR so we can review it next ?

Ladicek:
Rebased.

Ladicek:
Added one more commit that improves the replication test, although I'm not sure if that isn't really a deficiency of the replication client :shrug:

src/main/java/io/vertx/redis/client/RedisClusterConnectOptions.java
```diff
@@ -17,6 +17,7 @@
 
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.core.json.JsonObject;
+import io.vertx.core.tracing.TracingPolicy;
```
vietj, line 20: un-necessary import

src/main/java/io/vertx/redis/client/RedisConnectOptions.java
```diff
@@ -17,6 +17,7 @@
 
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.core.json.JsonObject;
+import io.vertx.core.tracing.TracingPolicy;
```
vietj, line 20: un-necessary import

src/main/java/io/vertx/redis/client/RedisSentinelConnectOptions.java
```diff
@@ -17,6 +17,7 @@
 
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.core.json.JsonObject;
+import io.vertx.core.tracing.TracingPolicy;
```
vietj, line 20: un-necessary import

src/main/java/io/vertx/redis/client/RedisStandaloneConnectOptions.java
```diff
@@ -17,6 +17,7 @@
 
 import io.vertx.codegen.annotations.DataObject;
 import io.vertx.core.json.JsonObject;
+import io.vertx.core.tracing.TracingPolicy;
```
vietj, line 20: un-necessary import

vietj (approved):
Minor import cleanup, besides that looks good to me

Ladicek:
Removed useless imports, thanks for noticing!

vietj:
@Ladicek you can merge and backport the PR

Ladicek:
Backported to 4.x in #398.

PR #400, opened 2023-07-16
Move blocking random UUID generation to executor thread to keep pipeline reactive

Arooba-git:
Hi! 👋 
Apparently the PoolOptions class has a blocking operation in its constructor that blocks the vertx event loop (as detected by BlockHound):
Screen Shot 2023-07-16 at 5 49 04 PM
This PR fixes the blocking code to ensure the event loop remains non-blocking.

### DIFF
--- a/src/main/java/io/vertx/redis/client/PoolOptions.java
+++ b/src/main/java/io/vertx/redis/client/PoolOptions.java
@@ -19,6 +19,8 @@ import io.vertx.codegen.annotations.DataObject;
 import io.vertx.core.json.JsonObject;
 
 import java.util.UUID;
+import java.util.concurrent.ExecutorService;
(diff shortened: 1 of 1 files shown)


vietj:
I think we already have soemthing similar for RNG generation in vertx-auth, perhaps we can reuse it ? cc @pmlopes

tsegismont:
@vietj we have several places in the Vert.x stack where we generate UUIDs, I don't believe it is necessary to use `VertxContextPRNG` for that.

@Arooba-git if you want to use blockhound, consider allowing some blocking calls
See https://github.com/reactor/BlockHound/blob/master/docs/customization.md

vietj:
@tsegismont we can discuss that next friday maybe

PR #402, opened 2023-08-18
Client-side caching support

craig-day:
Motivation:

Redis has some commands and server side support for clients to implement [client-side caching](https://redis.io/docs/manual/client-side-caching/) by using a key tracking approach. 

This proposes adding support by defining a new `CachingRedis` interface that extends `Redis`, along with a `RedisClientCache` interface to act as the local store. The client implementation will then checkout and maintain a long-lived connection that will act in subscribe mode to listen for invalidations. This is the approach required for the RESP2 protocol, but technically isn't required for RESP3 since connections can multiplex data and response types. However, I chose to keep using this approach to keep things simple by leveraging a long-lived connection as a read stream in subscribe mode.

I also included a simple implementation of a local cache store, an LRU. I'd be happy to exclude this if we want consumers to be forced to supply their own, but I started with something to make it usable out of the box.

Conformance:

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -24,6 +24,7 @@ Redis has a rich API and it can be organized in the following groups:
 * Strings - Commands to work with Strings.
 * Transactions - Commands to handle transaction lifecycle.
 * Streams - Commands to handle streaming.
+* Client-side caching - Commands to control client-side caching.
 
 == Using Vert.x-Redis
 
@@ -182,8 +183,53 @@ And from another place in the code publish messages to the queue:
 ----
 
 NOTE: It is important to remember that the commands `SUBSCRIBE`, `UNSUBSCRIBE`, `PSUBSCRIBE` and `PUNSUBSCRIBE` are `void`.
-This means that the result in case of success is `null` not a instance of response.
+This means that the result in case of success is `null` not an instance of response.
 All messages are then routed through the handler on the client.
+
+== Client-side Caching
+
+Redis supports client-side caching implementations using a strategy called _Tracking_.
+
+All modes of the client support caching except connections that are in pub/sub mode.
+
+To create a client with client-side caching, one would do:
+
+[source,$lang]
+----
+{@link examples.RedisExamples#clientCaching1}
+----
+
+A specific interface, `CachingRedis`, is exposed for a caching client that allows an invalidation handler to be attached or a flush command to be issued.
+The invalidation handler will be invoked with all the keys being invalidated whenever a message is received on the invalidations connection.
+
+To attach an invalidations handler:
+
+[source,$lang]
+----
+{@link examples.RedisExamples#clientCaching2}
+----
+
+To manually flush the client's cache store:
+
+[source,$lang]
+----
+{@link examples.RedisExamples#clientCaching3}
+----
+
+The client comes with a default cache store out of the box, but you can write your own if you prefer.
+
+The implementations are expected to follow the `ServiceLoader` conventions and all stores that are available at runtime from the classpath will be exposed.
+When more than 1 implementation is available the first one that can be instantiated and configured with success becomes the default.
+If none is available, then the default is a simple Least-Recently-Used (LRU) cache backed by a `LinkedHashMap`.
+
+[source,$lang]
+----
+{@link examples.RedisExamples#clientCaching4}
+----
+
+NOTE: The cache is not a write-through cache. A value will not be stored in the client-side cache until the value is fetched from Redis for the first time.
+To avoid write-then-read race conditions within the same batch, read commands that are part of a batch will not check the cache first.
+Additionally, the current implementation does not support the `OPTIN` or `NOLOOP` options.
 
 == Tracing commands
 
--- a/src/main/java/io/vertx/redis/client/CachingRedisOptions.java
+++ b/src/main/java/io/vertx/redis/client/CachingRedisOptions.java
@@ -1,0 +1,226 @@
+/*
+ * Copyright 2019 Red Hat, Inc.
+ * <p>
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * and Apache License v2.0 which accompanies this distribution.
+ * <p>
+ * The Eclipse Public License is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ * <p>
+ * The Apache License v2.0 is available at
+ * http://www.opensource.org/licenses/apache2.0.php
+ * <p>
+ * You may elect to redistribute this code under either of these licenses.
+ */
+package io.vertx.redis.client;
+
+import io.vertx.codegen.annotations.DataObject;
+import io.vertx.core.json.JsonObject;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+@DataObject(generateConverter = true)
+public class CachingRedisOptions {
+
+  /**
+   * The default number of max entries the cache will store.
+   */
+  public static final int DEFAULT_MAX_SIZE = 1024;
+
+  /**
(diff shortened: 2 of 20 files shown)


craig-day:
@pmlopes @vietj I opened this as a draft to get it on your radar, and I think the code is pretty much ready for review. I'm still working on adding additional test cases, but I don't plan to change the API or implementation except based on review feedback.

src/main/asciidoc/index.adoc
```diff
@@ -185,6 +186,30 @@ NOTE: It is important to remember that the commands `SUBSCRIBE`, `UNSUBSCRIBE`,
...
+{@link examples.RedisExamples#clientCaching1}
+----
+
+The default implementation will use a simple Least-Frequently-Used (LFU) cache backed by a `ConcurrentMap`.
+There is also simple Least-Recently-Used (LRU) cache implementation also backed by a `ConcurrentMap`.
+You can also provide your own implementation of the cache:
+
+[source,$lang]
+----
+{@link examples.RedisExamples#clientCaching1}
```
Ladicek, line 206:
```suggestion
{@link examples.RedisExamples#clientCaching2}
```

I think? :-)

src/main/java/io/vertx/redis/client/CachingRedisOptions.java
```diff
@@ -0,0 +1,217 @@
...
+  }
+
+  /**
+   * Create from JSON constructor.
+   *
+   * @param json the source JSON
+   */
+  public CachingRedisOptions(JsonObject json) {
+    this();
+    //CachingRedisOptionsConverter.fromJson(json, this);
```
Ladicek, line 77: ?

src/main/java/io/vertx/redis/client/impl/CachingRedisClientImpl.java
```diff
@@ -0,0 +1,261 @@
...
+            cache.put(cacheKey, resp);
+          }
+        })
+        .eventually(v ->
+          conn.close().onFailure(LOG::warn)));
+  }
+
+  @Override
+  public Future<List<@Nullable Response>> batch(List<Request> commands) {
+    // TODO: Do we want to support batch commands for the caching client?
```
Ladicek, line 133: This should probably be documented.

src/main/java/io/vertx/redis/client/impl/cache/LRUClientSideCache.java
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  @Override
+  protected long valueAfterGet(CacheKey key, long currentValue) {
+    return System.nanoTime();
+  }
+
+  @Override
+  protected long initialValueOnSet(CacheKey key, Response value) {
+    return System.nanoTime();
```
Ladicek, line 35: Should these be `System.currentTimeMillis()`, like the rest of the cache implementation code?

src/test/java/io/vertx/test/redis/CachingRedisClientTest.java
```diff
@@ -0,0 +1,125 @@
...
+          should.assertTrue(get2Response.result() == response);
+
+          api.send(Command.SET, key, "new value").onComplete(set2Response -> {
+            should.assertTrue(set2Response.succeeded());
+          });
+        });
+      });
+    });
+  }
+}
```
Ladicek, line 125: Could you add a `preservesContext` test, similarly to other Redis client tests?

Ladicek:
I don't feel qualified to review the caching code, but it seems like the 2 concurrent maps are assumed to be synchronized (_not_ the Java language `synchronized` keyword), even though they in fact don't necessarily have to be. Implementing a cache is hard.

One additional point. Not sure if this was intentional or not, but this seems kinda like a big deal: if you have a `CachingRedisClient`, call `connect()` on it, and use the resulting connection to perform operations, the cache will be completely bypassed. It seems to me that implementation-wise, caching should not be a _client_ wrapper, it should be a _connection_ wrapper. Caching on the client-level methods will come for free, as they just do `connect().send()` (or `connect().batch()`).

craig-day:
@Ladicek thanks for all of the points, and I also personally had the debate about client vs. connection. I decided to start here with the biggest target and use-case being for the `RedisAPI` interface. In my opinion, when you are calling `connect` and interacting directly with a connection, I feel like you've acknowledged that you don't necessarily get extra features. You just a connection to a server in which you can send commands, i.e. the lowest level option. Caching felt more like a higher level abstraction that wasn't necessarily explicit to the connection, so I started at the "client" layer. As long as the consumer uses `send()` or passes the client to `RedisAPI` (which uses `send`) then they get the benefits. It felt similar to connection pooling. If you want to leverage the pool, you can  just use `send` and the client will automatically checkout-send-release connections, but if you call `connect` then you have to deal with managing the sending and releasing. 

I do agree the most robust solution would be to directly integrate with the connection implementation to configure tracking based on options, but I wasn't sure where the invalidations listener and cache manager would live in such a case. Without going down the road of verticles or workers or using the event bus, it felt easier to make everything explicit and opt-in with a direct client. I could perhaps remove the `implements Redis` on the client, and then not expose `connect` so there isn't a confusing case where you skip the caching you explicitly opted in to.

src/main/java/io/vertx/redis/client/CachingRedisClient.java
```diff
@@ -0,0 +1,123 @@
...
+import io.vertx.redis.client.impl.cache.CacheKey;
+
+import java.util.Collection;
+
+/**
+ * A {@link Redis} client wrapper that implements client-side caching.
+ *
+ * @see <a href="https://redis.io/docs/manual/client-side-caching/">Client-side caching in Redis</a>
+ */
+public interface CachingRedisClient extends Redis {
```
vietj, line 31: should be named `CachingRedis` for consistency matters

craig-day: Done.

Ladicek:
@craig-day OK, that seems like a valid decision, but I guess it should be documented then?

craig-day:
@vietj @Ladicek I finally got back around to this. I've pushed a number of updates and I think this is ready for review. 

Some notes on the design choices. I chose to stick with a wrapper for a few reasons: 
1. We need somewhere to keep the long-lived invalidations connection. 
2. This same place also needs to have access to the local store to invalidate keys as it is notified. 
3. Additionally, there is some additional configuration needed to control how the client tracking gets configured when connections are opened.

I considered an alternative approach like creating a `CacheManager` that the client implementation could use, similar to the connection manager. The cache manager would maintain the long-lived connection and have access to the store, but to do this I would need to add an additional 2 arguments to each client constructor. One for the cache store, and another for the caching options. It would be nice to do this because we could then send the tracking configuration as the setup command for each connection, but since we only support a single setup command it means that this wouldn't work in the sentinel case since we already send a readonly command in some scenarios.

As an optimization, however, I did make the connections somewhat cache-aware. Each connection tracks whether it has had client tracking configured. This is done to avoid sending 2 commands for every command a client sends. Once a connection has had client tracking configured, it doesn't need to change unless it gets explicitly turned off. On that note, currently a client _could_ "break things" if they were to enable caching, checkout a connection, and then send the `CLIENT TRACKING off` command at some point. I don't know if we want to handle this explicitly, or even document it, but it is something that could happen.

src/test/java/io/vertx/test/redis/CachingRedisClientTest.java
craig-day: Added in `CachingRedisTest` and found a few bugs! :D

craig-day:
@Ladicek I think the last commit I just pushed is doing most of what you wanted. I removed the client creation away from the wrapper and into the default `Redis` methods. To do this I brought the `CachingRedisOptions` into `RedisOptions` in a similar way to the `PoolOptions` and `NetClientOptions` nested structures. Now to get caching you just enable the option. Doing this required a change in how custom cache implementations are provided, so that had to move to a `ServiceLoader` based pattern.

craig-day:
Closing as outdated

PR #406, opened 2023-09-20
[4.x] Adding equals/hashcode to members that compose key for connection pool

Ladicek:
Backport of #374 to 4.x. Without this, connection pooling basically doesn't work.


Ladicek:
The test failure is not caused by this PR, it fails on the current 4.x branch too (as well as in #407).

PR #407, opened 2023-09-20
[4.x] Fix how hash slot assignment is retrieved and stored

Ladicek:
Backport of #405 to 4.x.

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
@@ -16,6 +16,7 @@
 package io.vertx.redis.client.impl;
 
 import io.vertx.core.*;
+import io.vertx.core.impl.ContextInternal;
 import io.vertx.core.impl.logging.Logger;
 import io.vertx.core.impl.logging.LoggerFactory;
 import io.vertx.core.net.NetClientOptions;
@@ -30,6 +31,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Function;
 
 import static io.vertx.redis.client.Command.*;
@@ -124,6 +126,8 @@ public class RedisClusterClient extends BaseRedisClient implements Redis {
   private final RedisClusterConnectOptions connectOptions;
   private final PoolOptions poolOptions;
 
+  private final AtomicReference<Future<Slots>> slots = new AtomicReference<>();
+
   public RedisClusterClient(Vertx vertx, NetClientOptions tcpOptions, PoolOptions poolOptions, RedisClusterConnectOptions connectOptions, TracingPolicy tracingPolicy) {
     super(vertx, tcpOptions, poolOptions, connectOptions, tracingPolicy);
     this.connectOptions = connectOptions;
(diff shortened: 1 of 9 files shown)


Ladicek:
The test failure is not caused by this PR, it fails on the current 4.x branch too (as well as in #406).

vietj:
thanks @Ladicek let me fix the build and then you can rebase :-)

src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
```diff
@@ -220,31 +198,73 @@ private void connectionComplete(AtomicInteger counter, Slots slots, Map<String,
...
       } else {
-        onConnect.handle(Future.succeededFuture(new RedisClusterConnection(vertx, connectOptions, slots, connections)));
+        onConnected.handle(Future.succeededFuture(new RedisClusterConnection(vertx, connectOptions, slots,
+          () -> this.slots.set(null), connections)));
       }
     }
   }
 
-  private void getSlots(String endpoint, RedisConnection conn, Handler<AsyncResult<Slots>> onGetSlots) {
+  private Future<Slots> getSlots() {
```
vietj, line 209: instead here we should have a context

src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
```diff
@@ -220,31 +198,73 @@ private void connectionComplete(AtomicInteger counter, Slots slots, Map<String,
...
+    Slots slots = this.slots.get();
+    if (slots != null) {
+      return Future.succeededFuture(slots);
```
vietj, line 212: and use context.succeededFuture(slot)

src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
```diff
@@ -220,31 +198,73 @@ private void connectionComplete(AtomicInteger counter, Slots slots, Map<String,
...
+    }
+
+    Promise<Slots> promise = vertx.promise();
```
vietj, line 215: or context.promise()

src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java
```diff
@@ -139,72 +142,47 @@ public RedisClusterClient(Vertx vertx, NetClientOptions tcpOptions, PoolOptions
   @Override
   public Future<RedisConnection> connect() {
     final Promise<RedisConnection> promise = vertx.promise();
-    // attempt to load the slots from the first good endpoint
-    connect(connectOptions.getEndpoints(), 0, promise);
+    getSlots()
```
vietj, line 145: I think it can be replaced by getSlots().compose(slots -> ...) instead of onSuccess/onFailure

Ladicek: OK, I see how all the comments when put together make the code quite a bit nicer. Will do!

Ladicek: Actually no, we do need the extra `Promise` to pass to the `connect()` method, because that doesn't return a `Future`. It has a callback parameter, because it may retry and passing around the callback is cheaper than creating a `Future` for each retry attempt.

If we want to keep that, future combinators won't work. In which case, I think `onSuccess()` / `onFailure()` on the `Future` returned by `getSlots()` are just overall better.

Ladicek: For now, I at least pushed a commit that makes `getSlots()` use a proper context-bound promise.

PR #408, opened 2023-09-21
fix RedisOptions copy constructor

craig-day:
Motivation:

I kept running into `Redis waiting Queue is full` exceptions. After debugging, I found that when I copied the options during my setup, the `maxWaitingHandlers` was not copied and therefore got set to `0`.

Looks like this was inadvertently dropped when the options were decomposed into specific pool/connect options:
https://github.com/vert-x3/vertx-redis-client/pull/395/files#diff-c7335b2f2f7228ced4e9f5842327edab03b17586a270d7479f15df7eeb50f334L95

Conformance:

### DIFF
--- a/src/test/java/io/vertx/redis/client/test/RedisOptionsTest.java
+++ b/src/test/java/io/vertx/redis/client/test/RedisOptionsTest.java
@@ -1,6 +1,10 @@
 package io.vertx.redis.client.test;
 
+import io.vertx.core.tracing.TracingPolicy;
 import io.vertx.redis.client.RedisOptions;
+import io.vertx.redis.client.RedisRole;
+import java.util.ArrayList;
+import java.util.List;
 import org.junit.Test;
 
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


Ladicek (approved):

src/test/java/io/vertx/redis/client/test/RedisOptionsTest.java
```diff
@@ -1,6 +1,7 @@
 package io.vertx.redis.client.test;
 
 import io.vertx.redis.client.RedisOptions;
+import io.vertx.redis.client.RedisRole;
```
Ladicek, line 4: Seems useless?

Ladicek:
I think you need to rebase on latest `master` for the test failure to go away.

Also, this looks good to me, but I'll defer to @vietj (the original author of #395) to make the call. Maybe `maxWaitingHandlers` was supposed to be moved to `PoolOptions` or something like that, I'm not sure.

src/test/java/io/vertx/redis/client/test/RedisOptionsTest.java
craig-day: From a previous iteration of the test where I was setting options. Removed. Thanks!

holomekc:
Also ```TracingPolicy``` is missing in the copy constructor.

craig-day:
Thanks for the callout @holomekc, pushed that fix. 

@vietj @jponge can I get a review here?

src/test/java/io/vertx/redis/client/test/RedisOptionsTest.java
```diff
@@ -15,4 +19,24 @@ public void testRedisOptions() {
...
+
+  @Test
+  public void testOptionsCopy() {
+    List<String> endpoints = new ArrayList<>(3);
+    endpoints.add("redis://localhost:123");
+    endpoints.add("redis://localhost:124");
+    endpoints.add("redis://localhost:125");
+
+    // Set values for which there is no default to ensure they are copied correctly
+    RedisOptions original = new RedisOptions()
```
craig-day, line 31: Rather than setting every value, I set values which had no default. This allows the test to distinguish between a value not being copied and a value having no default. I could set a value for every option and assert on their copy, but I'm not sure if we see value in that. Happy to adjust this for any feedback.

PR #410, opened 2023-09-30
Feature/mutable options after refactor

holomekc:
Recreation after options refactoring. See #391


holomekc:
copy contstructor issue is already fixed here: #408 . I needed it though for my test.

holomekc:
Hi @vietj,

what do you think?

PR #412, opened 2023-10-11
Fix how hash slot assignment is retrieved and stored

### LINKED ISSUE #404: RedisClusterClient gives much higher load to first healthy node
https://github.com/vert-x3/vertx-redis-client/blob/9e88503327836e930ef616bbda8649c3ecfcf35b/src/main/java/io/vertx/redis/client/impl/RedisClusterClient.java#L140

For each command `RedisClusterClient.connect()` always calls `getSlots()` to send command `CLUSTER SLOTS` to first healthy node,  thus the first healthy node has very high CPU usage,  in my benchmark test with Quarkus,  the first healthy node consumes 80% CPU and other nodes consume 30% CPU,  this leads to extra latency and uneven load on Redis cluster.

The `getSlots()` call should be cached and automatically periodically updated, or updated by `MOVED` response from Redis cluster.

vietj:
Previously, the `RedisClusterClient` used to obtain the hash slot
assignment as the first step of each `connect()` call. This is fairly
expensive and increases the load on the first endpoint in the list
(because we target the first endpoint when issuing `CLUSTER SLOTS`).

It is also unnecessary. Redis always sends a redirection when the node
to which a command is sent is not assigned the hash slot targetted
by the command. Until we observe such redirection, the hash slot
assignment we observed before is still valid. Hence, we can store
the hash slot assignment in the `RedisClusterClient` and reuse it
for all `RedisClusterConnection` objects, until the `MOVED` error
is seen. In such case, we reset the hash slot assignment so that
the next `connect()` call fetches it again.

To avoid spurious failures (it could happen that the cluster is
resharded such that a command that would fail under the existing
hash slot assignment will no longer fail, because the hash slots
that were previously assigned to multiple nodes are now assigned
to a single node), the hash slot assignment is not kept permanently.
Instead, it is treated as a cache with configurable TTL, defaulting
to 1 second.

Fixes #404

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -141,14 +141,18 @@ To work with cluster the connection creation is quite similar:
 {@link examples.RedisExamples#example6}
 ----
 
-In this case the configuration requires one of more members of the cluster to be known. This list will be used to ask
-the cluster for the current configuration, which means if any of the listed members is not available it will be skipped.
+In this case the configuration requires one or more members of the cluster to be known.
+This list will be used to ask the cluster for the current configuration, which means if any of the listed members is not available it will be skipped.
 
-In cluster mode a connection is established to each node and special care is needed when executing commands. It is
-recommended to read redis manual in order to understand how clustering works. The client operating in this mode will do
-a best effort to identify which slot is used by the executed command in order to execute it on the right node. There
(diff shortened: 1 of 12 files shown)


Ladicek:
I thought you already backported the `equals`/`hashCode` fix in #411, so the 1st commit of this PR seems redundant?

PR #413, opened 2023-10-19
make Command.create() return existing instance for known commands

Ladicek:
When the `Command.create()` method is used to create a `Command` instance for a known command (not recommended, but also not forbidden), it used to return a generic instance that doesn't know anything about where the keys are in the command. Such generic instance is unusable with Redis cluster, because the target node will be selected randomly, not based on the key, and there's high chance such command will result in the `MOVED` redirect.

With this commit, `Command.create()` will return a pre-existing static instance for known commands, which is key-aware and works with Redis cluster.


Ladicek:
FTR, my main motivation is to fix https://github.com/quarkusio/quarkus/issues/28704, but I assume direct users of Vert.x Redis client can make the same mistake.

vietj:
thanks @Ladicek

PR #418, opened 2023-11-20
Load script on all master nodes in cluster

MichaelKubovic:
Motivation:

Redis uses command `SCRIPT LOAD` to load script to a script cache. This command should be run on every master node in cluster, otherwise node executing the script with `EVALSHA` fails on nodes that didn't load the script. Currently with clustered client, all `SCRIPT` commands are executed on a random node. My change handles this sub-command differently.

I have also considered a solution to introduce composite keys in `REDUCERS` (command + subcommand) but thought it would be premature abstraction, so I just implemented this directly without touching `REDUCERS`.

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisClusterConnection.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisClusterConnection.java
@@ -141,7 +141,7 @@ public class RedisClusterConnection implements RedisConnection {
         // can run anywhere
         if (REDUCERS.containsKey(cmd)) {
           sendToAllSlots(promise, req, cmd, args, forceMasterEndpoint, REDUCERS.get(cmd));
-        } else if(cmd.equals(SCRIPT) && Arrays.equals(args.get(0), "LOAD".getBytes())) {
+        } else if(cmd.equals(SCRIPT) && "LOAD".equalsIgnoreCase(new String(args.get(0)))) {
           sendToAllSlots(promise, req, cmd, args, forceMasterEndpoint, responses -> {
             // all nodes should compute the same sha
             assert responses.stream().map(Response::toString).collect(Collectors.toSet()).size() == 1;
--- a/src/main/java/io/vertx/redis/client/impl/RedisClusterConnection.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisClusterConnection.java
@@ -10,9 +10,11 @@ import io.vertx.redis.client.impl.types.ErrorType;
 
 import java.util.*;
 import java.util.function.Function;
+import java.util.stream.Collectors;
 
 import static io.vertx.redis.client.Command.ASKING;
 import static io.vertx.redis.client.Command.AUTH;
+import static io.vertx.redis.client.Command.SCRIPT;
 import static io.vertx.redis.client.Request.cmd;
 
 public class RedisClusterConnection implements RedisConnection {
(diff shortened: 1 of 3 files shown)


Ladicek:
Code LGTM, though it doesn't feel right to special case this particular command. But I don't have a better suggestion at the moment :-)

MichaelKubovic:
An alternative suggestion would be to expose API on `RedisClusterClient` to send an arbitrary command to all slots and let the code in the user-land deal with script loading. But I see two downsides to it: a) it exposes a low-level methods which would better be encapsulated b) high-level `RedisAPI::script()` will still suffer in cluster env.

Ladicek:
Oh I totally agree this should work out of the box, without reaching for any special API. We will need to do something similar to properly support Redis transactions in the cluster, so please don't take my comment above as a dismissal. It's just something we need to solve, one day. Not necessarily in this PR.

MichaelKubovic:
> We will need to do something similar to properly support Redis transactions in the cluster

We are running basic transactions just fine with a help of `batch`. You can't use `RedisAPI` methods unfortunately because they'd run each command on a different connection. The only thing that doesn't work with batching though is `WATCH`.

I could imagine a method on `RedisClusterClient` to create an object of `RedisAPIImpl` in a connection mode for a particular slot or just random node if not specified. Then you could use that instance to do transactions even without batch and with watch. But that's for another PR :)

Is there something I need to do besides waiting in order to get this reviewed by maintainers?

Ladicek:
When I mentioned transactions, I was thinking of something like this: https://github.com/quarkusio/quarkus/issues/32361#issuecomment-1759586485

src/main/java/io/vertx/redis/client/impl/RedisClusterConnection.java
```diff
@@ -138,23 +140,12 @@ public Future<Response> send(Request request) {
...
-
-          Future.all(responses).onComplete(composite -> {
-            if (composite.failed()) {
-              // means if one of the operations failed, then we can fail the handler
-              promise.fail(composite.cause());
-            } else {
-              promise.complete(REDUCERS.get(cmd).apply(composite.result().list()));
-            }
+          sendToAllSlots(promise, req, cmd, args, forceMasterEndpoint, REDUCERS.get(cmd));
+        } else if(cmd.equals(SCRIPT) && "LOAD".equalsIgnoreCase(new String(args.get(0)))) {
```
Ladicek, line 144:
```suggestion
        } else if (cmd.equals(SCRIPT) && args.size() > 1 && args.get(0).length == 4 && "LOAD".equalsIgnoreCase(new String(args.get(0)))) {
```

Nitpicking really, just for extra safety and to avoid needless allocation.

Ladicek:
I don't know why CI fails, it passed just fine in a PR of mine. @MichaelKubovic could you please rebase? That might help.

MichaelKubovic:
There are no new commits in master, so rebase didn't produce anything new. I've pushed a dummy commit to re-trigger build.

Ladicek:
You don't need to add an extra commit, you can just amend the commit on top with no changes -- that will change the commit date, which is enough.

But I see that didn't help. I don't know what's wrong :-/

MichaelKubovic:
There was a commit in master that fixes the issue :) Rebased.

MichaelKubovic:
That didn't help, so I tried to increase the container version. I run it locally with the updated version anyway + some more changes to the container setup because I can't use those ports on my MacOS.

MichaelKubovic:
Another attempt, it seems SCRIPT commands must be run on master nodes only.

Ladicek (approved):

Ladicek:
The `RedisCluster` API added in #466 / #467 allows executing arbitrary commands against all nodes / all master nodes.

PR #419, opened 2023-11-22
Redis cluster improvements

Ladicek:
Several commits, best reviewed in isolation:

- show all errors when connecting to Redis cluster fails
- use a subclass of ConnectException on connection failure in cluster/replication client
- fix caching of hash slot assignment in cluster client
- improve some code comments slightly
- treat sharded pub/sub commands as pub/sub commands
- improve how documentation links are generated in RedisAPI

### DIFF
--- a/src/main/java/io/vertx/redis/client/Command.java
+++ b/src/main/java/io/vertx/redis/client/Command.java
@@ -307,7 +307,7 @@ public interface Command {
   Command SORT = new CommandImpl("sort", -2, false, false, true, new KeyLocator(true, new BeginSearchIndex(1), new FindKeysRange(0, 1, 0)));
   Command SORT_RO = new CommandImpl("sort_ro", -2, true, false, true, new KeyLocator(true, new BeginSearchIndex(1), new FindKeysRange(0, 1, 0)));
   Command SPOP = new CommandImpl("spop", -2, false, false, false, new KeyLocator(false, new BeginSearchIndex(1), new FindKeysRange(0, 1, 0)));
-  Command SPUBLISH = new CommandImpl("spublish", 3, null, true, false, new KeyLocator(null, new BeginSearchIndex(1), new FindKeysRange(0, 1, 0)));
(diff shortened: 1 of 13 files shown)


Ladicek:
My main motivation for this is https://github.com/quarkusio/quarkus/issues/37041

luneo7 (approved):

sfali16 (commented):
Code changes look good. 

Some comments (1 and 2 are stylistic, and 3 might affect retries for non-cluster client connections):
1. Should the commit comment explaining the use of a class dervied from ConnectException be in the javadocs of one of the connect methods or the RedisClusterClient itself? 
2. Should we encapsulate this logic of using a RedisConnectException in its own function to better clarify the logic, and for future editors to reuse to do the right thing?

`onGotSlots.handle(Future.failedFuture(new RedisConnectException("Cannot connect to any of the provided endpoints"))); 
`
to 
```
onGotSlots.handle( failWithRetryableException( "Cannot connect to any of the provided endpoints"); 
// .. some other code 
/**
 * Return future that resolves with a RedisConnectException, an instance of ConnectException that will signal to clients that they can retry the connection
 */
public Future<RedisConnectException> failWithRetryableException( message ) { 
    return Future.failedFuture( new RedisConnectException( message ));
}
```

3. Finally should the RedisConnectException be returned by the RedisBaseClient ? Why is it only applicable to the cluster or replication clients?

Ladicek:
I could add something like `RedisConnectException.failedFuture(String)` and add a comment on that method. That's a good point.

It is required to use a `ConnectException` explicitly for the cluster and replication clients, because they fail the connecting process using a custom logic. The other clients propagate the original underlying exceptions directly, so they should be fine (I tested with the standalone client, but didn't bother checking with sentinel).

Ladicek:
I didn't do `RedisConnectException.failedFuture`, but I added a comment directly on `RedisConnectException`.

vietj (approved):
LGTM good work

luneo7 (approved):

PR #421, opened 2023-12-01
fix Redis connection sharing

### LINKED ISSUE #394: redis client shared in multiple verticles in high concurrency load will make redis call get response belonged to another request with some specific keys
### Questions

In my case, the Redis call seems to get a response that does not relate to the request but belonged to another request.

### Version

I'm writing a test case based on the redis-client newest commit 62c209b based on the master branch

### Context

Here is the test code: https://github.com/LangInteger/vertx-redis-client/blob/20230701_lang_test_shared_redis_connection/src/test/java/io/vertx/redis/RedisTestVerticle.java

Docker is needed to run the code.

### Steps to reproduce

1. clone the repo
2. open RedisTestVerticle
3. start the main function
4. you can see some Redis call success with some fail (in a weird way)

image

### Extra

Something I want to make a note:

- the redis key matters
  - I tested with another key pair `number_key` and `set_key`, and there will be no errors

    ```java
      private static final String REDIS_NUMBER_VALUE_KEY = "number_key";
      private static final String REDIS_SET_VALUE_KEY = "set_key";
    ```

- the number of verticles matters
  - I tested with verticle instance 1, and there will be no errors
  
    ```java
            vertx.deployVerticle(RedisTestVerticle.class,
              new DeploymentOptions().setInstances(1));
    ```

- the request load matters
  - I tested with for loop 10, and there will be no errors
  
    ```java
      public static void loopTest() {
        for (int i = 0; i < 10; i++) {
          System.out.println(Thread.currentThread() + "execute loop: " + i);
          test1();
        }
      }
    ```

Ladicek:
When a single `RedisConnection` is shared among multiple contexts (e.g. verticles), pipelining doesn't work correctly, because there is no coordination for concurrent access. This commit fixes that by making sure that all requests are performed from the same context on which the `NetSocket` was created. If the current context is different from the `NetSocket` context, the request is emitted on the correct context, which means no concurrent access, all requests are serialized.

Fixes #394

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisStandaloneConnection.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisStandaloneConnection.java
@@ -180,15 +180,21 @@ public class RedisStandaloneConnection implements RedisConnectionInternal, Parse
 
   @Override
   public Future<Response> send(final Request request) {
-    //System.out.println("send()#" + this.hashCode());
-    final Promise<Response> promise;
+    Promise<Response> promise = vertx.promise();
+    context.execute(() -> doSend(request, promise));
+    return promise.future();
+  }
 
+  private void doSend(final Request request, Promise<Response> promise) {
+    //System.out.println("send()#" + this.hashCode());
     if (closed) {
-      throw new IllegalStateException("Connection is closed");
+      promise.fail("Connection is closed");
+      return;
     }
 
     if (!((RequestImpl) request).valid()) {
-      return Future.failedFuture("Redis command is not valid, check https://redis.io/commands");
+      promise.fail("Redis command is not valid, check https://redis.io/commands: " + request);
+      return;
     }
 
     final CommandImpl cmd = (CommandImpl) request.command();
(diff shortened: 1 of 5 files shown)


Ladicek:
@vietj I believe this implements what we discussed earlier today. I'd really like your review on this :-)

When done, I will backport the fix to 4.x.

Ladicek:
Turns out when I moved from `NetClient.connect(SocketAddress)` to `NetClientInternal.connectInternal(ConnectOptions, Promise, ContextInternal)`, I didn't create the `ConnectOptions` fully -- I also have to pass the SSL options.

Fixed.

src/main/java/io/vertx/redis/client/impl/RedisStandaloneConnection.java
```diff
@@ -180,6 +180,22 @@ private void taintCheck(CommandImpl cmd) {
 
   @Override
   public Future<Response> send(final Request request) {
+    if (context == vertx.getContext()) {
```
vietj, line 183: I think instead we should use `ContextInternal.execute(Handler<Void>)` and modify `doSend` to use a `Promise<Response>` and have something like:

```java
public Future<Response> send(Request request) {
  Promise<Response> promise = vertx.getOrCreateContext().promise();
  context.execute(() -> doSend(request, promise));
  return promise.future();
}

private void doSend(Request request, Promise<Void>) { ... }
```

Ladicek: OK, that makes sense. Will fix.

Ladicek: Done.

src/test/java/io/vertx/test/redis/RedisClientTLSTest.java
```diff
@@ -149,7 +146,10 @@ public void testConnectionOptions(TestContext should) {
       .onSuccess(conn -> {
         conn.send(Request.cmd(Command.PING))
```
vietj, line 147: I think here we can simply do `onComplete(should.asyncAssertSuccess(res -> ...))` (applies to other parts)

Ladicek: I tried to stay away from rewriting this test too much, but I can try :-)

Ladicek: Done.

PR #428, opened 2024-02-06
Fix initialization of RedisOptions and Redis[*]ConnectOptions

Ladicek:
This commit makes sure that:

- all mandatory fields of `RedisOptions` are initialized (initializing them only in `Redis[*]ConnectOptions` is not enough due to the usage of copy constructors);
- all constructors of subclasses of `RedisConnectOptions` properly delegate to the corresponding superclass constructors;
- the `JsonObject`-accepting constructors of subclasses of `RedisConnectOptions` call the correct conversion function, after delegating to the corresponding superclass constructor, which calls its own conversion function;
- no information is lost during JSON serialization / deserialization, by reordering the getters / setters in the `Redis[Connect]Options` classes.

### DIFF
--- a/src/main/generated/io/vertx/redis/client/RedisConnectOptionsConverter.java
+++ b/src/main/generated/io/vertx/redis/client/RedisConnectOptionsConverter.java
@@ -40,16 +40,6 @@ public class RedisConnectOptionsConverter {
             obj.setPassword((String)member.getValue());
           }
           break;
-        case "endpoints":
-          if (member.getValue() instanceof JsonArray) {
-            java.util.ArrayList<java.lang.String> list =  new java.util.ArrayList<>();
-            ((Iterable<Object>)member.getValue()).forEach( item -> {
-              if (item instanceof String)
-                list.add((String)item);
-            });
-            obj.setEndpoints(list);
-          }
-          break;
         case "endpoint":
           break;
         case "connectionStrings":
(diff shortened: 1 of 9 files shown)


vietj:
should we have tests for this @Ladicek ?

Ladicek:
Added tests. Thanks for insisting on it, I actually discovered a few more subtle bugs -- most importantly, the JSON serialization/deserialization used to get/set the `endpoints` list multiple times, due to the existence of multiple getter/setter pairs, and that could actually lead to loss of information.

src/main/generated/io/vertx/redis/client/RedisConnectOptionsConverter.java
```diff
@@ -40,16 +40,6 @@ static void fromJson(Iterable<java.util.Map.Entry<String, Object>> json, RedisCo
             obj.setPassword((String)member.getValue());
           }
           break;
-        case "endpoints":
```
vietj, line 43: @Ladicek this still gets removed, is that a breaking change ?

Ladicek: It is not removed, it is _moved_, see below. This is to ensure that `endpoints` is serialized last and hence also deserialized last. This prevents information loss.

Previously, `endpoints` was serialized before `endpoint`, but `endpoint` is just the first element of the `endpoints` list. Since deserialization order follows serialization order, `endpoints` would be deserialized first and then `endpoint` would be deserialized, which leaves only a single element in the `endpoints` list. If there were multiple endpoints, only the first of them would be kept; the rest would be lost.

vietj: oh right, my bad :-)

vietj:
thank you @Ladicek

PR #429, opened 2024-02-06
[4.x] Fix initialization of RedisOptions and Redis[*]ConnectOptions

Ladicek:
This commit makes sure that:

- all mandatory fields of `RedisOptions` are initialized (initializing them only in `Redis[*]ConnectOptions` is not enough due to the usage of copy constructors);
- all constructors of subclasses of `RedisConnectOptions` properly delegate to the corresponding superclass constructors;
- the `JsonObject`-accepting constructors of subclasses of `RedisConnectOptions` call the correct conversion function, after delegating to the corresponding superclass constructor, which calls its own conversion function.

Backport of #428 to 4.x

### DIFF
--- a/src/main/java/io/vertx/redis/client/RedisClusterConnectOptions.java
+++ b/src/main/java/io/vertx/redis/client/RedisClusterConnectOptions.java
@@ -107,6 +107,11 @@ public class RedisClusterConnectOptions extends RedisConnectOptions {
   }
 
   @Override
+  public RedisClusterConnectOptions setPreferredProtocolVersion(ProtocolVersion preferredProtocolVersion) {
+    return (RedisClusterConnectOptions) super.setPreferredProtocolVersion(preferredProtocolVersion);
+  }
+
+  @Override
   public RedisClusterConnectOptions setPassword(String password) {
     return (RedisClusterConnectOptions) super.setPassword(password);
   }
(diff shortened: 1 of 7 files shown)


src/main/generated/io/vertx/redis/client/RedisOptionsConverter.java
```diff
@@ -20,24 +20,6 @@ public class RedisOptionsConverter {
    static void fromJson(Iterable<java.util.Map.Entry<String, Object>> json, RedisOptions obj) {
     for (java.util.Map.Entry<String, Object> member : json) {
       switch (member.getKey()) {
-        case "connectionString":
```
vietj, line 23: @Ladicek is that expected ?

vietj:
this seems to contain breaking change ? I can see that the java converters are changed

Ladicek:
Yes. I can drop the `@GenIgnore` annotations in this PR, which targets 4.x, because the Vert.x code generator processes properties in alphabetical order, therefore `getEndpoints()`/`setEndpoints()` always comes last and no information is lost.

In the corresponding PR to `master`, the `@GenIgnore` annotations actually prevent information loss, because the Vert.x codegen processes properties in declaration order. Actually, I can just reorder the getters and setters and don't have to add `@GenIgnore`.

I'll drop the `@GenIgnore` annotations here, and in the PR to `master`, I'll drop the `@GenIgnore` annotations and reorder the getters/setters.

Ladicek:
Done, both here and in #428.

Ladicek:
Rebased and resolved conflict.

PR #435, opened 2024-03-14
Improve javadoc of connection pool options

### LINKED ISSUE #288: Improve connection pool related RedisClientOptions and NetClientOptions parameter documentation
#### Describe the feature

The documentation regarding connection pooling is not sufficient and should be more detailed and explaining in my opinion.

#### Use cases

The configuration related to connection pooling supports the following parameters, please find my comments inline the code block:

```java
NetClientOptions netClientOptions = new NetClientOptions()
            .setConnectTimeout(connectTimeoutMs)

/* documentation says: Connection to be closed if no data is received in this time.
   It it what others call "socketTimeout"?
   How is the behavior related to the 'poolCleanerInterval' and the 'poolRecycleTimeout' of the RedisOptions? */
            .setIdleTimeout(idleTimeoutSeconds)
            .setIdleTimeoutUnit(TimeUnit.SECONDS)

/* Not redis-client specific, but it should be stated out that a connect operation lasts
   (connectTimeout + reconnectAttempts * ReconnectInterval);
   my first thought was that reconnect is about connection aborts after a connection was established */
            .setReconnectAttempts(reconnectAttempts)
            .setReconnectInterval(reconnectIntervalMs);

 Redis redisClient =Redis.createClient(
            get(),
            RedisOptions()

/* What does this setting do at all? Is it the interval, in which the 'poolRecycleTimeout' is check and stale connections are closed?
   Documentation says just "Tune how often in milliseconds should the connection pool cleaner execute."
   But this doesn't explain what it does and what that means. */
                .setPoolCleanerInterval(poolCleanerInterval)

/* This parameter is best explained in docs: "the timeout to keep an open connection on the pool waiting and then close"
   But: is this only relevant if also a 'poolCleanerInterval' is configured (defaults to -1 = inactive).
   And: how does this correlate/clash with the 'idleTimeout' of the NetClientOptions? */
                .setPoolRecycleTimeout(poolRecycleTimeoutMs)

                .setNetClientOptions(netClientOptions)
                .setMaxPoolSize(maxPoolSize)
                .setConnectionString(config.redisUrl);
```

#### Contribution

I would love to enhance the JavaDocs as well as the documentation by myself - as soon as I'm certain about these configuration options.

#### Credits

Thank you for this library, it's fun to use!

Ladicek:
Fixes #288

### DIFF
--- a/src/main/java/io/vertx/redis/client/PoolOptions.java
+++ b/src/main/java/io/vertx/redis/client/PoolOptions.java
@@ -54,8 +54,18 @@ public class PoolOptions {
   }
 
   /**
-   * Set a user defined pool name (for metrics reporting).
-   * @param name the user desired pool name.
+   * Get the connection pool name to be used for metrics reporting. The default name is a random UUID.
+   *
+   * @return the pool name
+   */
+  public String getName() {
+    return this.name;
+  }
+
+  /**
+   * Set the connection pool name to be used for metrics reporting. The default name is a random UUID.
+   *
+   * @param name the pool name
    * @return fluent self
    */
   public PoolOptions setName(String name) {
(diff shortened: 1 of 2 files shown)


Ladicek:
The _recycle timeout_ was probably meant to be a timeout after which an _unused_ connection is closed and evicted. However, the existing implementation doesn't do that -- it computes expiration time when the connection is created, and then never updates it. This PR just reflects reality.

It shouldn't be hard to fix that -- whenever the connection is returned back to the connection pool, the expiration time would have to be recomputed. That would be enough. I can do that in a 2nd commit in this PR, or separately. Any thoughts?

heubeck:
:heart_decoration:

PR #439, opened 2024-03-22
@GenIgnore non-canonical getters/setters of Redis[Connect]Options.endpoints

Ladicek:
The `RedisOptions` class has a field `endpoints` of type `List<String>` for which multiple accessors and mutators exist:

- `getEndpoints()`: the canonical getter
- `setEndpoints()`: the canonical setter
- `getEndpoint()`: returns the first element of the list
- `setEndpoint()`: turns the list into a single-element list
- `addEndpoint()`: adds one element to the list
- `setConnectionString()`: same as `setEndpoint()`
- `addConnectionString()`: same as `addEndpoint()`

The `RedisConnectOptions` class, which is `public` but is only used internally, has the same problem.

All these methods are treated as individual properties by the code generator that produces the JSON serialization/deserialization classes, even though they are all backed by a single field. Since the code generator considers properties in declaration order (in upcoming Vert.x 5; it is lexicographic order in Vert.x 4), we had to reorder the methods so that serialization and deserialization does not lose information.

That is arguably a provisional solution, as it is very much not obvious that declaration order matters. This commit provides a permanent solution: all the non-canonical methods are marked as `@GenIgnore`. Only the `getEndpoints()` and `setEndpoints()` methods are used for serialization/deserialization.

This is a breaking change, so it shall not be backported to Vert.x 4 and shall be documented as such.

### DIFF
--- a/src/main/generated/io/vertx/redis/client/RedisConnectOptionsConverter.java
+++ b/src/main/generated/io/vertx/redis/client/RedisConnectOptionsConverter.java
@@ -40,21 +40,6 @@ public class RedisConnectOptionsConverter {
             obj.setPassword((String)member.getValue());
           }
           break;
-        case "endpoint":
-          break;
-        case "connectionStrings":
-          if (member.getValue() instanceof JsonArray) {
-            ((Iterable<Object>)member.getValue()).forEach( item -> {
-              if (item instanceof String)
-                obj.addConnectionString((String)item);
-            });
-          }
-          break;
-        case "connectionString":
-          if (member.getValue() instanceof String) {
-            obj.setConnectionString((String)member.getValue());
-          }
-          break;
         case "endpoints":
           if (member.getValue() instanceof JsonArray) {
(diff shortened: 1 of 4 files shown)


Ladicek:
This is my proposal for https://github.com/vert-x3/wiki/wiki/5.0.0-Deprecations-and-breaking-changes:

## Vert.x Redis Client

### JSON serialization/deserialization of `RedisOptions`

The `RedisOptions` class has multiple methods to configure the Redis endpoints. These methods are still available, but the JSON format of this class has changed to only include one canonical field: `endpoints`.

If you rely on the JSON form of `RedisOptions` objects, note that the `endpoint`, `connectionString` and `connectionStrings` members of the JSON object are no longer recognized by the deserializer and are no longer generated by the serializer. Ensure that the necessary information is present in the **`endpoints`** member.

If you _do not_ rely on the JSON form of `RedisOptions` objects, this change does not affect you.

Issue #441, opened 2024-03-28
Expose more properties through CommandReporter


luneo7:
#### Describe the feature
Expose:
- `netSocket.remoteAddress().hostAddress()`
- `netSocket.remoteAddress().port()`
- `redisUri.socketAddress().host()`
-  `redisUri.socketAddress().port()`
As distinct properties through CommandReporter, and also make CommandReporter public alongside with the tags enum so we are able to map those in Quarkus without copying the enum there and also without doing string equals to see if the reporter is actually CommandReporter while instrumenting Redis with OpenTelemetry

#### Use cases

Use properties to properly create span attributes for OpenTelemetry tracing in Quarkus.

PR #442, opened 2024-03-28
Add more properties to CommandReporter

luneo7:
Add more properties to CommandReporter so we are able to add Server and Network attributes for OpenTelemetry tracing.

Closes https://github.com/vert-x3/vertx-redis-client/issues/441

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/CommandReporter.java
+++ b/src/main/java/io/vertx/redis/client/impl/CommandReporter.java
@@ -14,7 +14,11 @@ import java.util.function.Function;
 class CommandReporter {
   enum Tags {
     // Generic
+    NETWORK_PEER_ADDRESS("network.peer.address", reporter-> reporter.networkPeerAddress),
+    NETWORK_PEER_PORT("network.peer.port", reporter -> reporter.networkPeerPort),
     PEER_ADDRESS("peer.address", reporter -> reporter.address),
+    SERVER_ADDRESS("server.address", reporter -> reporter.serverAddress),
+    SERVER_PORT("server.port", reporter -> reporter.serverPort),
     SPAN_KIND("span.kind", reporter -> "client"),
 
     // DB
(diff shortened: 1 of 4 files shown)


luneo7:
@Ladicek here is my first take on the attributes that we need to properly construct the Server and Network attributes for OpenTelemetry. Can you take a look at it?

src/main/java/io/vertx/redis/client/impl/CommandReporter.java
```diff
@@ -11,8 +11,8 @@
 
 import java.util.function.Function;
 
-class CommandReporter {
-  enum Tags {
+public class CommandReporter {
+  public enum Tags {
```
Ladicek, line 15: This feels wrong, and this class is in the `impl` package, so it shouldn't be relied upon anyway.

src/main/java/io/vertx/redis/client/impl/CommandReporter.java
```diff
@@ -23,7 +23,11 @@ enum Tags {
     DB_USER("db.user", reporter -> reporter.user),
     DB_INSTANCE("db.instance", reporter -> reporter.database),
     DB_STATEMENT("db.statement", reporter -> reporter.command),
-    DB_TYPE("db.type", reporter -> "redis");
+    DB_TYPE("db.type", reporter -> "redis"),
+    NETWORK_PEER_ADDRESS("network.peer.address", reporter-> reporter.networkPeerAddress),
+    NETWORK_PEER_PORT("network.peer.port", reporter -> reporter.networkPeerPort),
+    SERVER_ADDRESS("server.address", reporter -> reporter.serverAddress),
+    SERVER_PORT("server.port", reporter -> reporter.serverPort);
```
Ladicek, line 30: Maybe move these above, to the _Generic_ section?

luneo7: Yup

Ladicek (approved):

Ladicek:
Thank you! You will also have to submit a copy of this PR against the `4.x` branch, if you want to use it anytime soon :-)

PR #444, opened 2024-04-05
 Add support for static topology configuration for the replication client @Ladicek

### LINKED ISSUE #343: Reading from read replicas when using Elasticache (cluster mode disabled)
#### Describe the feature

Currently, when setting the client to 'REPLICATION' mode, it performs an INFO command to 'master' and extracts its 'slaves' IPs from the reply. Using this with AWS Elasticache (cluster mode disabled) is faulty since the returned IPs are internal. This results in connection failures with the replicas leaving only communication with 'master'. This is not scalable and is not leveraging replication.

Using AWS Elasticache requires explicit configuration of the "Read endpoint" for leveraging the read replicas (see [here](https://docs.amazonaws.cn/en_us/AmazonElastiCache/latest/red-ug/Replication.Endpoints.html)).

The client should allow this configuration as an alternative for auto-discovering the replicas IPs.

#### Use cases

This feature will enable proper usage of vert.x redis client with AWS Elasticache (cluster mode disabled) which seems very important. In addition, this will benefit other Redis based servers which also hide their replicas IPs behind NATs

This can be easily reproduced by configuring the vert.x redis client to 'REPLICATION' mode and setting it to work with a basic Elasticache setup (cluster mode disabled) with 2 read replicas (configuring the endpoint to Elasticache 'primary' endpoint).

Ladicek:
Includes other changes, especially:

- introduced dedicated `RedisReplicationConnectOptions` (internal-only, but important anyway)
- fixed sending the `READONLY` command to replicas (only in the cluster mode, not in replication mode)
- documentation improvements
- testing improvements

Fixes #343


Ladicek:
The interesting thing about the testing improvements here is support for non-Linux OSs. I don't know if it works or not, I'd appreciate if anyone with Windows or macOS tested this (just `mvn clean verify`) and reported back.

PR #448, opened 2024-06-11
Fixed correct node selection for CLUSTER batch command

### LINKED ISSUE #447: Cluster batch command same slot calculation is wrong
When calculating slot for keys of batch command current implementations compare raw value of slot. This is wrong, it should calculate if command can be executed via endpoint that serves that slot. Example node 127.0.0.1:7000 will serve hash slot 0 - 5460. All commands that are in this hash slot should be executed without an issue via batch.

I have already written a fix and test for this, this issue is just for tracking.

pendula95:
Fixes #447

### DIFF
--- a/src/test/java/io/vertx/redis/client/test/RedisClusterTest.java
+++ b/src/test/java/io/vertx/redis/client/test/RedisClusterTest.java
@@ -4,6 +4,7 @@ import io.vertx.codegen.annotations.Nullable;
 import io.vertx.core.Context;
 import io.vertx.core.Future;
 import io.vertx.core.buffer.Buffer;
+import io.vertx.core.json.JsonObject;
 import io.vertx.ext.unit.Async;
 import io.vertx.ext.unit.TestContext;
 import io.vertx.ext.unit.junit.RunTestOnContext;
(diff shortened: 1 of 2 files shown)


src/test/java/io/vertx/redis/client/test/RedisClusterTest.java
```diff
@@ -1163,6 +1164,70 @@ public void setAndWaitEmptyBatch(TestContext should) {
...
+      String value = "value-" + i;
+      int endpoint;
+      int slot = ZModem.generate(key);
+      if (slot < 16384 / (options.getEndpoints().size() / 2)) {
+        endpoint = 0;
+      } else if (slot <= 2 * 16384 / (options.getEndpoints().size() / 2)) {
+        endpoint = 1;
+      } else {
+        endpoint = 2;
+      }
```
Ladicek, line 1184: This is likely to work, but seems fragile. Wouldn't it be best to connect to the actual cluster, read `CLUSTER SLOTS` (likely using the `Slots` class) and merge the commands based on the actual current assignment?

src/test/java/io/vertx/redis/client/test/RedisClusterTest.java
```diff
@@ -1163,6 +1164,70 @@ public void setAndWaitEmptyBatch(TestContext should) {
...
+
+    List<Request> rawCommands = new ArrayList<>();
+    for (int i = 0; i < 100000; i++) {
+      String key = "key-" + i;
+      String value = "value-" + i;
+      int endpoint;
+      int slot = ZModem.generate(key);
+      if (slot < 16384 / (options.getEndpoints().size() / 2)) {
+        rawCommands.add(Request.cmd(Command.SET).arg(key).arg(value));
+      }
```
Ladicek, line 1217: Same here.

src/test/java/io/vertx/redis/client/test/RedisClusterTest.java
pendula95: As main reason of using batch is for performance, waiting to execute a command just to perform batch seems counter intuitive. This should be done "offline" but I do agree this is fragile. I propose that we expose an additional helper function which would return endpoint of belonging key. This way we can use already known state of the cluster via Vert.x Cluster Redis APi

Ladicek: Not sure I follow. What I propose is to read `CLUSTER SLOTS` _once_, before you start to group commands into batches, so that the grouping is based on actual reality. (Although the actual reality in a newly formed cluster is likely to follow the static assignment in this PR.)

Exposing this as an API is an interesting idea, but it would expand the scope of this PR significantly.

pendula95: I just wanted to have a viable code in test that shows how to perform grouping via hashslot without first contacting redis to obtain slot distribution as this will introduce a delay. As this is not documented I wanted to have it in tests as most of the times that is where I look on how to use some functionality. As long as you know the number of masters this is viable for grouping. When batch is submitted Vert.x driver will lookup hash lost accurately and decide on the endpoint. But ok, to keep it clean I have submitted the modified tests.

Ladicek (approved):

Ladicek:
If you could submit a backport to the `4.x` branch, that would be great!

Issue #303, opened 2021-07-26
Create a higher-level Redis client wrapper that handles connection recovery


felixputera:
#### Read me

Read this first before creating an issue:

- do not use this issue tracker to ask questions, instead use one of these [channels](https://vertx.io/community/). Questions will likely be closed without notice.
- you shall create a feature request only when it is general purpose enough.
- make sure that the feature is not already

#### Describe the feature

This is suggested by @pmlopes in https://groups.google.com/g/vertx/c/vZZOFtFGOlI.

To quote his suggestion:

> Given that now we have a single interface for all the connection modes, single, sentinel, cluster and replication, perhaps a good contribution would be a wrapper helper that implements opinionated recovery semantics. For example, wrap the client, and register an exception handler that catches recoverable errors. In such events, queue all calls to "send" and "batch" until the underlying client is replaced with a new one.

#### Use cases

This will make the vertx redis client even easier to be used by end-users.

#### Contribution

Not decided yet.

felixputera:
As an aside, I know that Node.js Redis clients (like https://github.com/NodeRedis/node-redis and https://github.com/luin/ioredis) have this feature implemented too.

PR #459, opened 2024-08-06
Cluster and sentinel improvements

### LINKED ISSUE #320: Automatically refresh connections to redis cluster on shard/replica addition/deletion
#### Describe the feature

**Refresh connections to redis cluster whenever there is a change in cluster**

Current redis client implementation creates connections to redis cluster and uses them to send requests to redis cluster. However if redis cluster is changed (new shards/replicas are added/deleted) then it does not automatically refreshes the connection, due to which we see errors like `MOVED 7652 10.0.4.210:6379` and we have to redeploy our service.

If we could somehow detect changes in the cluster and refresh connections then we can get rid of deploying whenver we scale/downscale redis cluster

One simple solution I can think of is to save the cluster information when connecting to cluster for the first time using `CLUSTER INFO/CLUSTER SLOTS` command. Then periodically run this command and if there is any change in the cluster we can refresh the connections.

#### Use cases

- In today's world we upscale/downscale our data stores on the fly so it becomes cumbersome to redeploy micro-services after each scaling activity. This feature will remove the need to redeploy micro-services when scaling a redis cluster

#### Contribution

I am willing to contribute to this feature.

### LINKED ISSUE #372: Cluster mode (MOVED Exception)
### Questions

I'm using the RedisClusterClient which communicates with AWS elasticache cluster with 3 shards + 1 replica per shard.
While trying to write json value I'm getting this error:

```
 ERROR [com.del.san.SanityCheckLambda] (vert.x-eventloop-thread-1) MOVED 5776 x.x.x.x:6379
```

### Version

Which version(s) did you encounter this bug ?

4.3.7

Ladicek:
Fixes #320
Fixes #372

Related to #303

This will also fix the following Quarkus issue: https://github.com/quarkusio/quarkus/issues/40151

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -1,7 +1,7 @@
-= Vert.x-redis
+= Vert.x Redis
 :toc: left
 
-Vert.x-redis is redis client to be used with Vert.x.
+Vert.x Redis is a Redis client to be used with Vert.x.
 
 This module allows data to be saved, retrieved, searched for, and deleted in a Redis.
 Redis is an open source, advanced key-value store.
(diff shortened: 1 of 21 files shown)


Ladicek:
Documentation done and TODO in the test resolved.

PR #475, opened 2024-09-23
Add ability to provide `RedisConnectOptions` asynchronously

### LINKED ISSUE #356: Add support for AWS IAM
https://docs.aws.amazon.com/AmazonElastiCache/latest/red-ug/auth-iam.html

Technically we can use this mechanism as of today as we support redis authentication, we could however try to simplify the process to generating the v4 signature in a blocking call outside the event loop.

### LINKED ISSUE #364: Pick the protocol from the url if provided
Currently, the client negotiates the protocol version at connection level, or falls back to 2 if configured so.

Instead, we could pick it from the url if provided, as it is now a reserved property.

https://github.com/redis/redis-specifications/pull/12#pullrequestreview-1292235737

Ladicek:
In addition to that, this PR also adds `Redis[Connect]Options.getUser()` / `setUser()` and also support for the `protocol` query parameter in the Redis URI. Finally, this PR removes all the remaining usages of `Handler<AsyncResult<...>>` and replaces them with `Completable`.

Fixes #356
Closes #376
Closes #391
Closes #410
Fixes #364

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -80,7 +80,7 @@ In the configuration contains a `password` and/or a `select` database, these 2 c
 {@link examples.RedisExamples#example2}
 ----
 
-== Connecting using TLS
+=== Connecting using TLS
 
 You can connect to a Redis server using TLS by configuring the client TCP options, make sure to set:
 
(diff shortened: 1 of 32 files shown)


Ladicek:
Hi @holomekc, this should basically be the same as you did in #410. If you could take a look and see if this fixes your original issue, that would be great. Thanks!

Ladicek:
OK, thanks!

I kinda sorta co-maintain the Redis extension in Quarkus and indeed we'll need adjustments there as well.

PR #476, opened 2024-09-26
Use names specified by OpenTelemetry for tags

tsegismont:
See https://github.com/eclipse-vertx/vertx-tracing/issues/69

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/CommandReporter.java
+++ b/src/main/java/io/vertx/redis/client/impl/CommandReporter.java
@@ -22,12 +22,12 @@ class CommandReporter {
     SPAN_KIND("span.kind", reporter -> "client"),
 
     // DB
-    // See https://github.com/opentracing/specification/blob/master/semantic_conventions.md
+    // See https://opentelemetry.io/docs/specs/semconv/database/redis/
 
     DB_USER("db.user", reporter -> reporter.user),
-    DB_INSTANCE("db.instance", reporter -> reporter.database),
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/redis/client/impl/CommandReporter.java
```diff
@@ -22,12 +22,12 @@ enum Tags {
...
     // DB
-    // See https://github.com/opentracing/specification/blob/master/semantic_conventions.md
+    // See https://opentelemetry.io/docs/specs/semconv/database/redis/
 
     DB_USER("db.user", reporter -> reporter.user),
-    DB_INSTANCE("db.instance", reporter -> reporter.database),
-    DB_STATEMENT("db.statement", reporter -> reporter.command),
-    DB_TYPE("db.type", reporter -> "redis");
+    DB_NAMESPACE("db.namespace", reporter -> reporter.database),
+    DB_QUERY_TEXT("db.query.text", reporter -> reporter.command),
```
Ladicek, line 29: I think this is actually `db.operation.name`, per the OpenTelemetry docs.

Ladicek:
LGTM otherwise.

src/main/java/io/vertx/redis/client/impl/CommandReporter.java
tsegismont: Good catch, thanks @Ladicek 

I assumed we had the full command here

Ladicek (approved):

PR #478, opened 2024-10-04
Extract the Redis subscription -> Vert.x event bus handler

### LINKED ISSUE #291: Handle Subscribe/Unsubscribe response messages
## Describe the feature
Other kind of messages apart from "message" are not being handled by default. Currently only "messages" are being sent as Json objects to the even bus, for others a warning message is being printed. e.g: WARNING: No handler waiting for message: [subscribe, channel, 1]

Unless this is the expected behaviour I think all kind of messages should be handled in the same way.

All kind of responses are described [here](https://redis.io/topics/pubsub)

> Format of pushed messages
> A message is a Array reply with three elements.
> 
> The first element is the kind of message:
> 
> subscribe: means that we successfully subscribed to the channel given as the second element in the reply. The third argument represents the number of channels we are currently subscribed to.
> 
> unsubscribe: means that we successfully unsubscribed from the channel given as second element in the reply. The third argument represents the number of channels we are currently subscribed to. When the last argument is zero, we are no longer subscribed to any channel, and the client can issue any kind of Redis command as we are outside the Pub/Sub state.
> 
> message: it is a message received as result of a PUBLISH command issued by another client. The second element is the name of the originating channel, and the third argument is the actual message payload.

#### Use cases

A list of use cases this feature will enable and the value it creates.

#### Contribution

I would be happy to add a pr, please check this commit: https://github.com/cristian-com/vertx-redis-client/commit/47a5b1a1392c1177a0412d6ef91dbe2e159bae77

If you consider the above applies I will make it look nicer in a private method and handle the missing cases for unsubscribe responses .

This issue was originally mentioned at: https://github.com/quarkusio/quarkus/issues/13887#issuecomment-746941729

Ladicek:
The `EventBusHandler` does almost the same job, but has to be set on the `RedisConnection` explicitly (using the `handler()` method). In addition to forwarding the messages to the event bus, the new `EventBusHandler` also forwards subscriptions/unsubscriptions.

Fixes #291


vietj:
thanks @Ladicek for stepping up on this decoupling

PR #479, opened 2024-10-18
initial support for single-node Redis transactions in cluster

Ladicek:


Ladicek:
Draft because at least documentation is required. I might want to add a few more tests, too.

Ladicek:
Should be ready now.

Ladicek:
This could be backported to 4.x in the future.

PR #485, opened 2024-12-06
Improve Redis cluster

Ladicek:
- Make `Request`, `Response` and `Command` data objects
- Change `groupByNodes` return type and make `RedisCluter` `@VertxGen`
- Add documentation for `RedisCluster`

### DIFF
--- a/src/main/asciidoc/index.adoc
+++ b/src/main/asciidoc/index.adoc
@@ -278,6 +278,36 @@ When the cache is empty, the first attempt to acquire a connection will execute 
 The cache has a configurable TTL (time to live), which defaults to 1 second.
 The cache is also cleared whenever any command executed by the client receives the `MOVED` redirection.
 
+=== Cluster Utilities
+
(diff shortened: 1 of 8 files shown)


Ladicek:
Hi @vietj, @tsegismont, @pendula95, here are the changes we discussed earlier today.

Ladicek:
Note to self: there's a breaking change here, so need to add a note into https://github.com/vert-x3/wiki/wiki/5.0.0-Deprecations-and-breaking-changes#vertx-redis-client after this is merged.

Ladicek:
I made a quick test with `vertx-rx` and indeed `Request`, `Response` and `Command` are no longer generated, while `RedisCluster` newly is. Which seems right on the first sight, but I'm no expert here.

tsegismont (approved):
The API changes look good to me. Thank you @Ladicek !

When this is merged, don't forget to update the breaking changes page as well as the migration guide.

vietj:
thanks @Ladicek

vietj (commented):
LGTM

PR #494, opened 2025-04-16
Allow using only a username in Redis URL authinfo

joschi:
The `io.vertx.redis.client.impl.RedisURI` class parsed URIs with only a username in the URI userinfo component but silently dropped it afterwards.

`redis://my-user@example.com:6379/0` was accepted without throwing an exception but the user defaulted to `default` or alternatively to the given `user` query parameter.

This change set modifies the behavior so that the username from the example URI above (`my-user`) is being properly used and that the password is still unset.

Please note that `redis://my-user@example.com` (password is unset/null) and `redis://my-user:@example.com` (password is empty string) are different and both cover valid connection cases.

Additionally the Javadoc for `RedisURI` has been extended to cover the valid URI schems and query parameters.

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisURI.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisURI.java
@@ -28,7 +28,7 @@ import java.util.Map;
 
 /**
  * Utility to parse Redis URLs. The URI should follow the general scheme:
- * {@code redis://[username:password@][host][:port][/[database]}
+ * {@code redis://[[username][:password]@][host][:port][/[database]]}
  *
  * <p>An example URI can be found at <a href="https://redis.io/topics/rediscli">Redis cli docs</a>.</p>
  *
--- a/src/main/java/io/vertx/redis/client/impl/RedisURI.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisURI.java
@@ -27,14 +27,42 @@ import java.util.HashMap;
 import java.util.Map;
 
 /**
- * Utility to parse redis URLs. The URI should follow the scheme: redis://[username:password@][host][:port][/[database]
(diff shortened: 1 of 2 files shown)


src/main/java/io/vertx/redis/client/impl/RedisURI.java
```diff
@@ -27,14 +27,42 @@
 import java.util.Map;
 
 /**
- * Utility to parse redis URLs. The URI should follow the scheme: redis://[username:password@][host][:port][/[database]
- * An example URI can be found at <a href="https://redis.io/topics/rediscli">Redis cli docs</a>
+ * Utility to parse Redis URLs. The URI should follow the general scheme:
+ * {@code redis://[username:password@][host][:port][/[database]}
```
Ladicek, line 31: The change seems reasonable, but then this pattern needs to be adjusted. I believe it should look like this now:

```suggestion
 * {@code redis://[[username][:password]@][host][:port][/[database]]}
```

src/main/java/io/vertx/redis/client/impl/RedisURI.java
```diff
@@ -27,14 +27,42 @@
...
+ * <p>An example URI can be found at <a href="https://redis.io/topics/rediscli">Redis cli docs</a>.</p>
+ *
+ * <p>Valid URI schemes are: {@code redis} (TCP), {@code rediss} (TCP+TLS), {@code unix} (UNIX socket).</p>
+ *
+ * <p>
+ *   Additionally the following query parameters are supported (while the dedicated URI components take precedence):
+ *   <ul>
+ *     <li>{@code db}: database</li>
+ *     <li>{@code user}: username</li>
+ *     <li>{@code password}: username</li>
```
Ladicek, line 42:
```suggestion
 *     <li>{@code password}: password</li>
```

Ladicek (approved):

Issue #502, opened 2025-08-05
RESP3 parser does not handle "bulk string" (binary data) keys properly in map responses


dburyak:
### Version

5.0.1

### Context

# Description
`MultiType` class design assumes that RESP3 `map` type responses from redis must always contain valid UTF-8 **keys**. While redis docs (and actual behavior) clearly say that keys in `map` responses may be of any type including BulkStrings (binary) (https://redis.io/docs/latest/develop/reference/protocol-spec/#maps).

So if you try to retrieve maps with binary keys (redis HASH in the reproducer project), negative bytes will get replaced with the same UTF codepoint `65533` (question mark). I.e. keys gets scrambled, it's impossible to restore original received binary data (which is correct by the way in the received buffer) from the `MultiType` object returned to the user of redis-client.

So there are two inter-related parts of this bug:
 - `MultiType` class does not support binary keys by design
 - binary data gets scrambled when it's being decoded as UTF8 string, here: https://github.com/vert-x3/vertx-redis-client/blob/a9aed026cd3c09848af7b8ef7631a63a45e928f1/src/main/java/io/vertx/redis/client/impl/types/MultiType.java#L88

See attached reproducer project for more details.

I'd like to fix this myself, please let me know if it's fine for me to contribute.

# Workarounds
I didn't find any direct workaround since the root cause sits very deep inside the parser. If you already have HASHES in redis with binary fields, then it seems that it's impossible to read them with vertx-redis client (with `HGETALL` of course). And this can be extrapolated to any commands+data-structures that have `map` resp3 response types.

I can see only 2 indirect and limited workarounds though:
 - avoid using binary keys in cases where you have control over it
 - store binary keys as base64 encoded strings, if we have control over the writing process

### Steps to reproduce

(see reproducer project, it contains reproducible easy to run out-of-the-box test)
1. Create `HASH` redis entry with binary fields (keys). Values are irrelevant for this bug. Binary keys should contain negative bytes (bytes greater than 127 if you interpret them as unsigned ones).
2. Call `HGETALL` using vertx-redis-client.
3. There's no direct support to get **keys** of returned response object as binary. Only as UTF8 strings.
4. There's no way to reverse UTF8 string keys back into bytes.

### Do you have a reproducer?

https://github.com/dburyak/vertx-redis-resp3-map-keys-decoding-bug

PR #538, opened 2025-12-18
502 map bin keys

dburyak:
Implements #502 . Adds support for binary keys in MultiType. Specifically, this is needed for reading binary fields (keys) of HASH-es with HGETALL command.

The approach is to always store elements of multi in `multi` array, and evaluate corresponding map with string or/and binary keys when either is requested. This approach makes this class NOT thread safe, but it is not a really useful feature since under typical usage scenarios there won't be concurrent access. Hence, it's better to keep this class simple and not-thread-safe, and if the need to use it concurrently arises in some exotic use cases, then let the callers deal with it externally.

NOTE: I had to decide whether to keep using un-ordered HashMap for binary keys too, or change both strMap and binMap to be LinkedHashMap. In fact, Redis preserves order of fields in HASH-es, and relevant commands like HKEYS, HVALS, HSCAN and HGETALL particularly, return values in their respective order. With this change, the existing method `MultiType#getKeys()` will also preserve order, which was not the case due to usage of HashMap. Even though technically this is a change of behavior, it can not affect any existing users, as they were unable to rely on the order anyway prior to this change.

### DIFF
--- a/src/main/java/io/vertx/redis/client/Response.java
+++ b/src/main/java/io/vertx/redis/client/Response.java
@@ -278,7 +278,53 @@ public interface Response extends Iterable<Response> {
   }
 
   /**
-   * Get this size of this multi response.
+   * Get this Multi response value at a binary key.
+   * <p>
+   * This is a variant of {@link #get(String)} that accepts binary keys. Use
+   * this method when the map contains binary keys that may not be interpreted
+   * as valid UTF-8 chars. For example, redis HASH with binary field names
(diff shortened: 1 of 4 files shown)


Ladicek:
Rebased and squashed, and also made some tiny changes. I'm not especially fond of how `MultiType.toString()` changes behavior based on previous calls, but I couldn't figure out a better way to do this, so let's stick with that for now.

PR #553, opened 2026-04-14
add support for Redis 8 and Valkey

Ladicek:
I tried my best to ensure there's no compatibility breakage, but the generated files are so huge I'm not entirely sure. I'll try to run a proper test later (e.g. Revapi), before merging this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,3 +8,8 @@ The Vert.x Redis client provides an asynchronous API to interact with a [Redis](
 ## Documentation:
 
 The documentation can be read from the official vert.x [Documentation](http://vertx.io/docs/vertx-redis-client/java/).
+
+## Valkey
+
+This client also supports [Valkey](https://valkey.io/).
+Use the `valkey://` or `valkeys://` connection string scheme and the `ValkeyAPI` interface, which only includes commands supported by Valkey.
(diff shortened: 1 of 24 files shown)


Ladicek:
Revapi is quite crazy in how it includes types in its analysis, but [japicmp](https://siom79.github.io/japicmp/) works well, and it doesn't detect any breakage here. I actually removed the backward compatibility overloads I added, and it did detect that, so it's not entirely stupid :-)

Ladicek:
Moved back to Revapi, because it actually caught the intentional difference of `@GenIgnore`  (which affects how auto-generated code looks like). Added `revapi.json` with the list. It includes a weird case of a non-public API that is _sometimes_ reported by Revapi (and sometimes _not_ reported :exploding_head:).

geny200:
@Ladicek Hi, does this mean `vertx-redis-client` will officially supports `Valkey`, and future protocol changes introduced in `Valkey`?

Ladicek:
I certainly hope that Valkey is _not_ going to make protocol changes. I don't mind maintaining a list of supported commands (since it's generated automatically...), but if they ever want to make a change in the protocol, this falls down super fast. I guess I should add a note that this is still primarily a Redis client and Valkey stays supported as long as it stays on the Redis protocol.

geny200:
Thx for answer)

>  I guess I should add a note that this is still primarily a Redis client and Valkey stays supported as long as it stays on the Redis protocol.

Yes; I think that note should be in readme/docs.

PR #563, opened 2026-05-29
Simple test for Unix Domain Socket connectivity

### LINKED ISSUE #562: Check connectivity with Unix Domain Sockets
The client documentation says it's possible to connect with a Unix Domain Socket but we don't have a test for this.

The test should be skipped when the runtime version is not 16+ (we don't need to verify that Vert.x can use a native transport with older Java versions).

tsegismont:
Closes #562

### DIFF
--- a/src/test/java/io/vertx/tests/redis/client/RedisUnixDomainSocketTest.java
+++ b/src/test/java/io/vertx/tests/redis/client/RedisUnixDomainSocketTest.java
@@ -9,6 +9,7 @@ import io.vertx.redis.client.RedisOptions;
 import io.vertx.redis.client.Request;
 import io.vertx.tests.redis.containers.RedisStandalone;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledForJreRange;
 import org.junit.jupiter.api.condition.EnabledOnOs;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.api.extension.RegisterExtension;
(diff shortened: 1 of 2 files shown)


src/test/java/io/vertx/tests/redis/client/RedisUnixDomainSocketTest.java
```diff
@@ -0,0 +1,50 @@
...
+import org.testcontainers.junit.jupiter.Container;
+import org.testcontainers.junit.jupiter.Testcontainers;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.condition.OS.LINUX;
+
+@ExtendWith(VertxExtension.class)
+@Testcontainers
+@EnabledOnOs({LINUX})
```
Ladicek, line 26: I think you also need `@EnabledForJreRange(min = JRE.JAVA_16)` here.

tsegismont: Ooops, thanks!

Ladicek:
Test failed because of https://status.redhat.com/incidents/jn6r256zc62c, we'll have to wait until that is resolved.

Ladicek (approved):

Ladicek:
I think we want this in the 5.1 branch too. Let me backport.

tsegismont:
Thanks!

PR #564, opened 2026-05-29
Close socket when init of connection fails

### LINKED ISSUE #555: RedisStandaloneConnection and underlying NetSocket leak when HELLO/AUTH/SELECT/setup fails during connection initialization
**Affected versions:** 4.5.23 (verified in production); same code present in 4.5.26 and on the `4.x`, `5.0`, and `master` branches as of today.

## Summary

When `RedisConnectionManager$RedisConnectionProvider.init(...)` runs the handshake (`HELLO`, optional fallback `AUTH`, `SELECT`, user setup) and any step fails with a Redis-protocol error rather than a TCP-level disconnect, the failed `Future` is propagated to the pool but the underlying `NetSocket` is never closed and the partially-constructed `RedisStandaloneConnection` is never told to `forceClose()`. The pool's `ConnectFailed` action then runs `Remove`, decrementing the pool capacity counter and freeing the slot — but the leaked socket and connection object remain alive (referenced by `netSocket.handler(new RESPParser(connection, ...))`, `closeHandler(connection::end)`, `exceptionHandler(connection::fail)`).

Because Redis returns AUTH errors as in-protocol error replies (`-WRONGPASS …`) without closing the TCP connection, `closeHandler` never fires and nothing eventually reaps the leak. Each subsequent failed handshake leaks one more socket. The leak is invisible to `pool-recycle-timeout` and `pool-cleaner-interval` because those iterate `((RedisEndpoint) e).pool` — the leaked connection never made it into the pool's slot array.

## Reproduction

1. Stand up a Redis (or ElastiCache-compatible) server that returns `WRONGPASS` for some valid-looking credential.
2. Configure the Vert.x Redis client with `password` matching the broken credential.
3. Loop: `redis.send(Request.cmd(Command.GET).arg("k"))`. Each call triggers `BaseRedisClient.send → connect → pool.acquire → connector.connect → init → hello → AUTH-fail`.
4. Observe via `jcmd <pid> GC.class_histogram | grep -E 'RedisStandaloneConnection|NetSocketImpl'` that both counts grow linearly with call count, regardless of `max-pool-size`.
5. `ss -tn dport = :6379 state established | wc -l` shows the matching number of OS sockets.

## Production evidence

Two pods of the same image, AWS ElastiCache (Redis 7) with IAM auth, after a cluster reload window where AUTH transiently fails:

| Pod | Container age | `RedisStandaloneConnection` instances | OS sockets to :6379 |
| --- | --- | --- | --- |
| A | 19h | 3,996 | 7,857 |
| B | 3h12m | 4,647 | 9,214 |

Heap also showed exactly one `RedisConnectionManager`, one `RedisEndpoint`, one `ConnectionKey`, one `RedisAPIImpl`, and 4,647 `NetSocketImpl` instances 1:1 with the standalone connections. All sockets were `ESTABLISHED` at the OS level — no `FIN_WAIT_*`, no `CLOSE_WAIT` — proving the JVM had not initiated socket close.

## Root cause (4.5.23 source)

`src/main/java/io/vertx/redis/client/impl/RedisConnectionManager.java`, lines 215–243 in `init(...)`. Three failure callbacks (`hello.failed()`, `select.failed()`, `setupResult.failed()`) each call `ctx.execute(ctx.failedFuture(...), onConnect)` and return without closing `netSocket` or invoking `connection.forceClose()`. Same shape exists on `master` (lines 226–237; uses `Future.compose` chain but still has no `.onFailure(...)` cleanup hook).

## Suggested fix

Either (a) attach `.onFailure(err -> connection.forceClose())` to the composed `init()` future, or (b) add explicit `connection.forceClose()` calls before each `ctx.execute(failedFuture..., onConnect)` in 4.x. Both will trigger `netSocket.close()` (via `RedisStandaloneConnection.forceClose() → netSocket.close()` at line 89), which fires `closeHandler` → `connection::end`, releasing all references.

Happy to send a PR.

pendula95:
Fixes #555 

I was able to make a test that proves the point but only for AUTH-fail. Cases where  SELECT and upgradeSSL still do not have the test. Even the test that was written is more of an integration test and I should try to make one with mocked NetServer.

### DIFF
--- a/src/main/java/io/vertx/redis/client/impl/RedisConnectionManager.java
+++ b/src/main/java/io/vertx/redis/client/impl/RedisConnectionManager.java
@@ -194,7 +194,8 @@ public class RedisConnectionManager implements Function<RedisConnectionManager.C
               // must upgrade protocol
               return so
                 .upgradeToSsl()
-                .compose(v -> init(ctx, options, so, listener));
+                .compose(v -> init(ctx, options, so, listener))
+                .onFailure(ex-> so.close());
             } else {
               // no need to upgrade
               return init(ctx, options, so, listener);
@@ -227,14 +228,17 @@ public class RedisConnectionManager implements Function<RedisConnectionManager.C
         .compose(hello -> {
           // perform select
           return select(ctx, connection, redisURI.select());
-        }).compose(select -> {
+        })
+        .compose(select -> {
           // perform setup
           return setup(ctx, connection, setup);
-        }).map(setup -> {
+        })
+        .map(setup -> {
           // connection is valid
           connection.setValid();
-          return new ConnectResult<>(connection, 1, 0);
-        });
+          return new ConnectResult<>((RedisConnectionInternal) connection, 1, 0);
+        })
+        .onFailure(err -> connection.forceClose());
     }
 
     private Future<Void> hello(ContextInternal ctx, RedisStandaloneConnection connection, RedisURI redisURI, RedisConnectOptions options) {
--- a/src/test/java/io/vertx/tests/redis/client/RedisClientConnectionFailTest.java
+++ b/src/test/java/io/vertx/tests/redis/client/RedisClientConnectionFailTest.java
@@ -1,0 +1,104 @@
+package io.vertx.tests.redis.client;
+
+import io.vertx.core.Future;
+import io.vertx.junit5.RunTestOnContext;
+import io.vertx.junit5.VertxExtension;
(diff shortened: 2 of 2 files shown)


pendula95:
I have added test for upgradeSSL, I am happy with the current state of the tests. Introducing a mock netserver for testing this might introduce hard to maintain test with changes in the RESP protocol. Sticking to test containers seems like an ok option.

PR #568, opened 2026-06-08
Make Redis Resource shareable

### LINKED ISSUE #451: Make Redis Resource shareable
Like other resources add option to share Redis. Example PG pool sharing https://vertx.io/docs/vertx-pg-client/java/#_pool_sharing or JDBC https://vertx.io/docs/vertx-jdbc-client/java/#_using_default_shared_data_source

pendula95:
Resolves #451 

I guess I should make a 5.0 version as well? As this is a new feature I don't think a backport to 4x is needed?

### DIFF
--- a/src/main/java/examples/RedisExamples.java
+++ b/src/main/java/examples/RedisExamples.java
@@ -309,6 +309,38 @@ public class RedisExamples {
     options.setClusterTransactions(RedisClusterTransactions.SINGLE_NODE);
   }
 
+  public void example17(Vertx vertx) {
+    Redis redis = Redis.createClient(
+      vertx,
+      new RedisOptions()
+        .setConnectionString("redis://localhost:6379")
+        .setShared(true)
+        .setPoolName("my-redis"));
+    vertx.deployVerticle(() -> new VerticleBase() {
+      @Override
+      public Future<?> start() {
+        return redis.send(Request.cmd(Command.PING)).mapEmpty();
+      }
+    }, new DeploymentOptions().setInstances(4));
+  }
+
+  public void example18(Vertx vertx) {
+    vertx.deployVerticle(() -> new VerticleBase() {
+      Redis redis;
+
+      @Override
+      public Future<?> start() {
+        redis = Redis.createClient(
+          vertx,
+          new RedisOptions()
+            .setConnectionString("redis://localhost:6379")
+            .setShared(true)
+            .setPoolName("my-redis"));
+        return Future.succeededFuture();
+      }
+    }, new DeploymentOptions().setInstances(4));
+  }
+
   public void tracing1(RedisOptions options) {
     options.setTracingPolicy(TracingPolicy.ALWAYS);
   }
--- a/src/main/java/io/vertx/redis/client/PoolOptions.java
+++ b/src/main/java/io/vertx/redis/client/PoolOptions.java
@@ -19,13 +19,22 @@ import io.vertx.codegen.annotations.DataObject;
 import io.vertx.codegen.json.annotations.JsonGen;
 import io.vertx.core.json.JsonObject;
 
-import java.util.UUID;
-
 @DataObject
 @JsonGen(publicConverter = false)
 public class PoolOptions {
 
+  /**
+   * Default shared pool config = {@code false}
+   */
+  public static final boolean DEFAULT_SHARED_POOL = false;
+
+  /**
+   * Actual name of anonymous shared pool = {@code __vertx.DEFAULT}
+   */
+  public static final String DEFAULT_NAME = "__vertx.DEFAULT";
+
   private String name;
+  private boolean shared;
   private int cleanerInterval;
   private int maxSize;
   private int maxWaiting;
@@ -33,7 +42,8 @@ public class PoolOptions {
   private long maxLifetime;
 
   public PoolOptions() {
-    name = UUID.randomUUID().toString();
+    name = DEFAULT_NAME;
+    shared = DEFAULT_SHARED_POOL;
     // thumb guess based on web browser defaults
     cleanerInterval = 30_000;
     maxSize = 6;
@@ -44,10 +54,12 @@ public class PoolOptions {
 
   public PoolOptions(PoolOptions other) {
     this.name = other.name;
+    this.shared = other.shared;
     this.cleanerInterval = other.cleanerInterval;
     this.maxSize = other.maxSize;
     this.maxWaiting = other.maxWaiting;
     this.recycleTimeout = other.recycleTimeout;
+    this.maxLifetime = other.maxLifetime;
   }
 
   public PoolOptions(JsonObject json) {
@@ -56,7 +68,7 @@ public class PoolOptions {
   }
 
   /**
-   * Get the connection pool name to be used for metrics reporting. The default name is a random UUID.
+   * Get the connection pool name to be used for metrics reporting.
    *
    * @return the pool name
    */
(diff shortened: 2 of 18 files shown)


vietj:
5.0 will not get releases anymore, we should not also backport features to 4.5.x

src/main/java/examples/RedisExamples.java
```diff
@@ -309,6 +309,38 @@ public void example16(RedisOptions options) {
...
+          vertx,
+          new RedisOptions()
+            .setConnectionString("redis://localhost:6379")
+            .setShared(true)
+            .setPoolName("my-redis"));
+        return Future.succeededFuture();
+      }
+    }, new DeploymentOptions().setInstances(4));
+  }
+
```
Ladicek, line 343: I'm sorry I don't understand sharing, so I'll have a couple dumb questions here :-)

I can understand `example17`, where the client is basically shared between whoever grabs the variable.

I'm not sure about `example18`. Is it only shared between all instances of that one verticle? Noone else can access it.

src/main/java/io/vertx/redis/client/impl/CloseableRedis.java
```diff
@@ -0,0 +1,54 @@
...
+  public Future<RedisConnection> connect() {
+    return delegate.connect();
+  }
+
+  @Override
+  public Future<Void> close() {
+    ContextInternal closingCtx = vertx.getOrCreateContext();
+    Promise<Void> promise = closingCtx.promise();
+    closeFuture.close(promise);
+    return promise.future();
```
Ladicek, line 42: Shouldn't this method also delegate (somehow?) to `delegate.close()`?

src/main/java/io/vertx/redis/client/PoolOptions.java
```diff
@@ -44,10 +54,12 @@ public PoolOptions() {
 
   public PoolOptions(PoolOptions other) {
     this.name = other.name;
+    this.shared = other.shared;
     this.cleanerInterval = other.cleanerInterval;
     this.maxSize = other.maxSize;
     this.maxWaiting = other.maxWaiting;
     this.recycleTimeout = other.recycleTimeout;
+    this.maxLifetime = other.maxLifetime;
```
Ladicek, line 62: Thanks for the fix!

src/test/java/io/vertx/tests/redis/client/SharedRedisClientTest.java
```diff
@@ -0,0 +1,152 @@
...
+    Future.all(client1.close(), client2.close()).onComplete(test.succeedingThenComplete());
+  }
+
+  @Test
+  public void testExplicitCloseReleasesLeaseOnly(VertxTestContext test) {
+    Redis client1 = Redis.createClient(context.vertx(), standaloneOptions("lease-release-pool"));
+    Redis client2 = Redis.createClient(context.vertx(), standaloneOptions("lease-release-pool"));
+
+    client1.close()
+      .compose(v -> client2.send(Request.cmd(Command.PING)))
```
Ladicek: We should assert that this command successfully responds I guess.

Ladicek:
Again, I don't know _anything_ about resource sharing, so my review isn't probably worth much. One more question, besides the comments: is reusing the existing _pool name_ configuration for resource sharing potentially problematic? IIUC, the existing pool name is only used for metrics, but now, it's also used for sharing.

src/main/java/io/vertx/redis/client/impl/CloseableRedis.java
pendula95: Nope, this would result in immediate close without respecting other consumers. 
Release lease; when last lease is gone, owner closeFuture runs → connectionManager shuts down

Ladicek: Ah I see. It's in the `BaseRedisClient` constructor. I'll add a comment there.

src/main/java/io/vertx/redis/client/impl/BaseRedisClient.java
```diff
@@ -26,11 +27,12 @@ public abstract class BaseRedisClient<OPTS extends RedisConnectOptions> implemen
   protected final Supplier<Future<OPTS>> connectOptions;
   protected final RedisConnectionManager connectionManager;
 
-  public BaseRedisClient(Vertx vertx, NetClientOptions tcpOptions, PoolOptions poolOptions, Supplier<Future<OPTS>> connectOptions, TracingPolicy tracingPolicy) {
+  public BaseRedisClient(Vertx vertx, NetClientOptions tcpOptions, PoolOptions poolOptions, Supplier<Future<OPTS>> connectOptions, TracingPolicy tracingPolicy, CloseFuture closeFuture) {
     this.vertx = (VertxInternal) vertx;
     this.connectOptions = connectOptions;
     this.connectionManager = new RedisConnectionManager(this.vertx, tcpOptions, poolOptions, (Supplier) connectOptions, tracingPolicy);
     this.connectionManager.start();
+    closeFuture.add(completion -> connectionManager.close().onComplete(completion));
```
Ladicek: I see this is how closing in case of sharing is done. The callback is basically a copy of the `close()` method, I wonder if it shouldn't just call `close()` instead?

src/main/java/examples/RedisExamples.java
pendula95: Most of the implementation together with the documentation was inspired by PG pool sharing. 
https://vertx.io/docs/vertx-pg-client/java/#_pool_sharing

In the example18 any verticle that was spawned by the same vertx object will reuse the same client as long as the name matches and shared option is set.

src/test/java/io/vertx/tests/redis/client/SharedRedisClientTest.java
```diff
@@ -0,0 +1,155 @@
...
+    Redis client1 = Redis.createClient(context.vertx(), standaloneOptions("lease-release-pool"));
+    Redis client2 = Redis.createClient(context.vertx(), standaloneOptions("lease-release-pool"));
+
+    client1.close()
+      .compose(v -> client2.send(Request.cmd(Command.PING))
+        .onSuccess(rest -> {
+          test.verify(() -> assertEquals("PONG", rest.toString()));
+        }).onFailure(test::failNow))
+      .eventually(client2::close)
+      .onComplete(test.succeedingThenComplete());
```
Ladicek, line 111: I think this should just be


```suggestion
    client1.close()
      .compose(v -> client2.send(Request.cmd(Command.PING))
      .map(resp -> {
          assertEquals("PONG", rest.toString());
          return resp;
      })
      .compose(v -> client2.close())
      .onComplete(test.succeedingThenComplete());
```

This is a linear pipeline that is easy to understand. What you did doesn't establish order between asserting on the PING command response and the `eventually()` callback, which is wrong.

src/main/java/examples/RedisExamples.java
Ladicek: Hmm, I see. Well, I don't get the documentation then, but the PG documentation is the same, so whatever :-)

src/test/java/io/vertx/tests/redis/client/SharedRedisClientTest.java
pendula95: Thanks

src/main/java/io/vertx/redis/client/impl/BaseRedisClient.java
pendula95: You are right

Ladicek:
One of the added tests failed. Since it wasn't changed in the last commits, the test might be flaky.

pendula95:
> One of the added tests failed. Since it wasn't changed in the last commits, the test might be flaky.

I have changed the test so that it also fails more stable. Root cause of fail was that INFO clients in cluster mode also includes inter cluster connections that cluster nodes have. To exclude this I have switched to use CLIENT list

Ladicek (approved):

Ladicek:
I don't know anything about resource sharing or `CloseFuture`, @vietj could you please take a look at this too?

PR #571, opened 2026-06-18
Add client self-identification via CLIENT SETINFO

vchomakov:
# Add client self-identification via CLIENT SETINFO

## What
After the handshake, the client reports itself to the server via `CLIENT SETINFO` (Redis 7.2+):
- `lib-name` = `vertx-redis-client`, `lib-ver` = version from the JAR
- Enabled by default; opt out with `setClientIdentification(false)`
- Upstream frameworks can append suffixes (`vertx-redis-client(suffix1;suffix2)`) via `addLibrarySuffix(...)`

```java
new RedisOptions().addLibrarySuffix("quarkus-redis_v3.x");
new RedisOptions().setClientIdentification(false); // opt out
```

## Design
Pipelined after `HELLO` (single round-trip), per the Redis spec.
Graceful degradation: all `CLIENT SETINFO` failures are ignored, so pre-7.2 servers and restrictive ACLs (`NOPERM`) still connect.
Suffix validation: setters throw on illegal chars (whitespace ( ) ;); handshake also drops invalid ones as a safety net.
New  `clientIdentification` / `librarySuffixes` options round-trip via JSON.

## Scope
Suffixes are passed as opaque strings; a typed name/version API may follow.

## Testing
L1 unit: name composition + suffix validation
L2 (Redis 6.2): connects when CLIENT SETINFO is unsupported
L3 (Redis 7.2): lib-name / lib-ver / suffix / disable verified via CLIENT INFO

### DIFF
--- a/src/test/java/io/vertx/tests/redis/internal/RedisClientVersionTest.java
+++ b/src/test/java/io/vertx/tests/redis/internal/RedisClientVersionTest.java
@@ -67,7 +67,7 @@ public class RedisClientVersionTest {
   public void testIsValidSuffixRejectsNonAscii() {
     // control characters and DEL
     assertFalse(RedisClientVersion.isValidSuffix("has\ttab"));
-    assertFalse(RedisClientVersion.isValidSuffix("hasdel"));
+    assertFalse(RedisClientVersion.isValidSuffix("has\u007Fdel"));
     // non-ASCII BMP characters
     assertFalse(RedisClientVersion.isValidSuffix("café"));
     // characters outside the Basic Multilingual Plane (encoded as a surrogate pair)
--- a/src/test/java/io/vertx/tests/redis/internal/RedisClientVersionTest.java
+++ b/src/test/java/io/vertx/tests/redis/internal/RedisClientVersionTest.java
@@ -1,0 +1,60 @@
+package io.vertx.tests.redis.internal;
+
+import io.vertx.redis.client.impl.RedisClientVersion;
+import org.junit.jupiter.api.Test;
+
+import java.util.Arrays;
+import java.util.Collections;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class RedisClientVersionTest {
+
+  @Test
+  public void testLibraryName() {
+    assertEquals("vertx-redis-client", RedisClientVersion.NAME);
+  }
+
+  @Test
+  public void testLibraryVersionIsPresent() {
+    assertNotNull(RedisClientVersion.VERSION);
+    assertFalse(RedisClientVersion.VERSION.isEmpty());
+  }
+
+  @Test
+  public void testFormatLibraryNameWithoutSuffixes() {
+    assertEquals("vertx-redis-client", RedisClientVersion.formatLibraryName(null));
+    assertEquals("vertx-redis-client", RedisClientVersion.formatLibraryName(Collections.emptyList()));
+  }
+
+  @Test
+  public void testFormatLibraryNameWithSingleSuffix() {
+    assertEquals("vertx-redis-client(quarkus-redis_v3.x)",
+      RedisClientVersion.formatLibraryName(Collections.singletonList("quarkus-redis_v3.x")));
+  }
+
+  @Test
+  public void testFormatLibraryNameWithMultipleSuffixes() {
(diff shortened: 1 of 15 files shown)


Ladicek (changes requested):
Looks good! I've got two ideas:

I think the library suffixes should be more clearly tied to client ID. How about we used the following method names on the config instead?

| your current name | my suggestion |
| --- | --- |
| `isClientIdentification` | `isClientId` |
| `setClientIdentification` | `setClientId` |
| `getLibrarySuffixes` | `getClientIdSuffixes` |
| `setLibrarySuffixes` | `setClientIdSuffixes` |
| `addLibrarySuffix` | `addClientIdSuffix` |

Also, the suffix validation is buggy when it comes to code points outside the BMP. But I'm thinking we should perhaps limit the suffix to ASCII characters anyway. We could just check if the `char` is `> 32` and `< 127` (and isn't `(`, `)` and `;`). WDYT? (The regexps at https://redis.io/docs/latest/commands/client-setinfo/ look really weird, so I'm not sure if I can take anything from that...)

Finally, we probably shouldn't mention Quarkus in the documentation or the javadoc. No strong reason, it just feels wrong :-)

Ladicek:
Also, it looks like you didn't sign the ECA yet. See https://www.eclipse.org/legal/eca/ for more info. I don't know if you have a corporate agreement with Eclipse or how that works, sorry -- but I'm sure there's someone on your side who knows.

vchomakov:
> Looks good! I've got two ideas:
> 
> I think the library suffixes should be more clearly tied to client ID. How about we used the following method names on the config instead?
> 
> your current name	my suggestion
> `isClientIdentification`	`isClientId`
> `setClientIdentification`	`setClientId`
> `getLibrarySuffixes`	`getClientIdSuffixes`
> `setLibrarySuffixes`	`setClientIdSuffixes`
> `addLibrarySuffix`	`addClientIdSuffix`
> Also, the suffix validation is buggy when it comes to code points outside the BMP. But I'm thinking we should perhaps limit the suffix to ASCII characters anyway. We could just check if the `char` is `> 32` and `< 127` (and isn't `(`, `)` and `;`). WDYT? (The regexps at https://redis.io/docs/latest/commands/client-setinfo/ look really weird, so I'm not sure if I can take anything from that...)
> 
> Finally, we probably shouldn't mention Quarkus in the documentation or the javadoc. No strong reason, it just feels wrong :-)

Thanks, @Ladicek, for the review! I've addressed your comments in 960282a:

- **Naming** — renamed the config methods to tie them to client ID: `isClientId` / `setClientId`, `getClientIdSuffixes` / `setClientIdSuffixes`, and `addClientIdSuffix` (JSON property names updated to match).
- **Suffix validation** — switched to your suggested ASCII rule (`> 32 && < 127`, excluding `(`, `)`, `;`), which also fixes the code-points-outside-the-BMP case. Added tests covering control chars, a non-ASCII BMP char, and a surrogate pair.
- **`lib-ver`** — when the version can't be resolved from the JAR manifest, it's now omitted entirely instead of reporting `unknown`.
- **Docs** — dropped the Quarkus mentions from the documentation and javadoc.

Let me know if you'd like any further changes!

src/test/java/io/vertx/tests/redis/internal/RedisClientVersionTest.java
```diff
@@ -57,4 +62,21 @@ public void testIsValidSuffix() {
     assertFalse(RedisClientVersion.isValidSuffix("has;semi"));
     assertFalse(RedisClientVersion.isValidSuffix(null));
   }
+
+  @Test
+  public void testIsValidSuffixRejectsNonAscii() {
+    // control characters and DEL
+    assertFalse(RedisClientVersion.isValidSuffix("has\ttab"));
+    assertFalse(RedisClientVersion.isValidSuffix("hasdel"));
```
Ladicek: Does this line contain the DEL char (127)? I didn't look outside GitHub, so I'm not 100% sure, but it would feel better if we used an escape sequence here (`\u007F`).

Ladicek:
Looks great, just one tiny comment. I'm not approving yet just to avoid accidental merge. I'll approve once you get the ECA sorted out. Thanks!

src/test/java/io/vertx/tests/redis/internal/RedisClientVersionTest.java
vchomakov: Yes, switched to using `\u007F` per your suggestion.

vchomakov:
> Looks great, just one tiny comment. I'm not approving yet just to avoid accidental merge. I'll approve once you get the ECA sorted out. Thanks!

Thanks, @Ladicek , ECA is now signed!

Ladicek (approved):

Ladicek:
Thanks!

PR #581, opened 2026-09-08
Increase RedisSentinelMasterFailoverTest timeout

tsegismont:
The RedisSentinelMasterFailoverTest was failing with a timeout after 30 seconds.

The test performs two consecutive Redis master failovers, each using retryUntilSuccess() with 50 retries at 500ms intervals.


tsegismont:
@Ladicek I thought I had sent a PR back in July, but it seems I didn't :sweat:

Ladicek (approved):

